PR #9, opened 2014-04-24
Extend splunk modules to allow cluster role configuration

hyclak:
Please forgive me if these changes are the wrong direction in terms of best practices, etc. I was unable to find how to do exactly what I wanted. As things stand now, both the splunk and splunk::foo classes need to be included in a node definition for the role to be properly configured. It's been a while since I've touched Puppet, so I am most definitely rusty.

This request exposes the necessary class parameters to create a splunk cluster with a master indexer, additional index peers and search heads. This allows the class to be used in ENC tools fairly easily. 

Please review, comment and/or provide any suggestions.

Thanks!

Matt

### DIFF
--- a/Modulefile
+++ b/Modulefile
@@ -1,6 +1,6 @@
 name    'puppetlabs-splunk'
-version '3.0.1'
-source  'https://github.com/puppetlabs/puppetlabs-splunk.git'
+version '3.0.1-1'
+source  'https://github.com/hyclak/puppetlabs-splunk.git'
 author  'Puppet Labs'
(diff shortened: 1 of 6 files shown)


tphoney:
Thanks for the large amount of work you put in. There are a number of things that need to be carried out before it will be considered for merging. Can you please squash your commits, add some tests, update the documentation so the puppet users know how to use the new parameters.
Thanks again.

tphoney:
Thanks for the work, i feel that another PR is better suited. Closing this in favour of #15.

PR #17, opened 2015-07-09
Repository and password changes

ckyriakidou:
Created a way of installing splunk if there are apt or yum repositories involved (not just from plain package)
Also added a default way for adding username/password information accross many systems for splunkforwarder.
Sponsored by Baloise Group


lib/puppet/provider/splunk_output/ini_setting.rb
```diff
@@ -7,9 +7,9 @@
   def self.file_path
     case Facter.value(:osfamily)
     when 'windows'
-      'C:\Program Files\Splunk\etc\system\local\outputs.conf'
+      'C:\Program Files\Splunk\etc\apps\search\local\outputs.conf'
```
hunner: What inspired this change?

manifests/forwarder.pp
```diff
@@ -49,24 +49,30 @@
...
 ) inherits splunk::params {
-  include staging
 
   $virtual_service = $splunk::params::forwarder_service
-  $staged_package  = staging_parse($package_source)
   $staging_subdir  = $splunk::params::staging_subdir
 
   $path_delimiter  = $splunk::params::path_delimiter
+  if $pkg_provider != undef and $pkg_provider != 'yum' and  $pkg_provider != 'apt' {
+  include staging
```
hunner: Could you fix the indentation here?

manifests/forwarder.pp
```diff
@@ -75,26 +81,13 @@
...
-    value   => "${server}_${logging_port}",
-    tag     => 'splunk_forwarder',
-  }
-  splunkforwarder_output { 'defaultgroup_server':
-    section => "tcpout:${server}_${logging_port}",
-    setting => 'server',
-    value   => "${server}:${logging_port}",
-    tag     => 'splunk_forwarder',
-  }
+   create_resources( 'splunkforwarder_input', hiera_hash('splunk::forwarder::forwarder_input',$::splunk::forwarder::forwarder_input))
```
hunner: We would prefer that hiera function calls not happen inside public modules, as many users don't have hiera and this will cause errors. Could you add parameters to pass in the forwarder_input and output values so that users can pass in a hash directly, use the defaults as before, or call hiera_hash to pass in the hash?

manifests/params.pp
```diff
@@ -130,6 +141,27 @@
...
+  }
+  $forwarder_input = {
+    'default_host' => {
+      section      => 'default',
+      setting      => 'host',
+      value        => "$::clientcert",
+    }
+
+  }
+  $pass4SymmKey = '$1$WZmW8GzCIbYB'
```
hunner: Where is this variable used?

lib/puppet/provider/splunk_output/ini_setting.rb
ckyriakidou: This one was to match inputs.conf which when I use splunk forwarder-add <splunk-server> it puts the forwarder in search\local\inputs.conf so I thought that  was the right file. I could just change it for the inputs.conf though and revert everythign else back to system if that would work?

manifests/password.pp
```diff
@@ -0,0 +1,57 @@
...
+# sponsored by balgroup
+class splunk::password( $password_config_file = $splunk::params::password_config_file,
+                        $secret_file          = $splunk::params::secret_file,
+                        $secret               = $splunk::params::secret,
+                        $password_content     = $splunk::params::password_content,
+                        $service_password     = 'splunk_forwarder',
+                        $virtual_service      = $splunk::params::forwarder_service,
+                        $license              = 'license_splunkforwarder',
+                      ) inherits splunk::params {
+ if ! defined(Class['splunk::forwarder']) and ! defined(Class['splunk']){
```
hunner: Looks like this whole class block needs indentation fixed

lib/puppet/provider/splunkforwarder_input/ini_setting.rb
```diff
@@ -7,9 +7,11 @@
   def self.file_path
     case Facter.value(:osfamily)
     when 'windows'
-      'C:\Program Files\SplunkUniversalForwarder\etc\system\local\inputs.conf'
+      FileUtils::mkdir_p 'C:\Program Files\SplunkUniversalForwarder\etc\apps\search\local'
```
hunner: I really don't love mkdir -p being called every single time the provider is run. Why manage this directory here instead of in a manifest? Does the package installation not handle it?

Worst case scenario, it should only mkdir -p if the directory doesn't exist, but even that would be discouraged as it is a system change that would happen even on --noop puppet runs...

hunner:
Could you also squash the commits?

lib/puppet/provider/splunk_output/ini_setting.rb
hunner: http://docs.splunk.com/Documentation/Splunk/6.2.3/Forwarding/Configureforwarderswithoutputs.confd#Custom_versions indicates that both of these paths are correct under some conditions and that the difference in the path is determined "on the type of forwarder and other factors."

Additionally, it says that "it is recommended that you work with just a single copy of the file, which you place in $SPLUNK_HOME/etc/system/local/."

As I don't know how splunk works and don't use it, it would be helpful if you could read that page and let me know what the best decision would be.

manifests/forwarder.pp
hunner: Also, it would be best if the default hash value for the parameter resulted in the above thee input/outputs so that it remains backwards-compatible.

manifests/forwarder.pp
```diff
@@ -107,7 +100,7 @@
   # If the purge parameters have been set, remove all unmanaged entries from
   # the inputs.conf and outputs.conf files, respectively.
   if $purge_inputs  {
-    resources { 'splunkforwarder_input':  purge => true; }
+   resources { 'splunkforwarder_input':  purge => true; }
```
hunner: This should be 2-space indentation.

manifests/forwarder.pp
ckyriakidou: will add this in params.pp so that it's overwriteable in forwarder.pp.

PR #19, opened 2015-07-20
Release 3.2.0

hunner:


bmjen:
There should probably be upper bounds to the dependencies in the metadata.json.

PR #25, opened 2015-09-14
config items are collected with tag - make sure it exists

asquelt:
`splunk_forwarder` tag is used for dependency setting (package, service), so it's essential all configuration items created by `splunk::forwarder` class contains those tags (unless user deliberately decides to override them).


hunner:
Instead of adding tags to the `create_resources` of `splunk_forwarder` could you just change the collectors `<| tag == 'splunk_forwarder' |>` below to `<| tag == 'splunk::forwarder' |>`

hunner:
Actually, this is probably fine as-is

PR #26, opened 2015-12-10
added technology addon feature

crayfishx:
The splunk module has no support for configuring forwarders for TA's (technology addons) which, according to the docs, have their input forwarders configured in $SPLUNK_HOME/etc/apps/<appname>/local/inputs.conf

I'm using a fork of this module to be able to have this feature - it currently only supports configuring inputs and installing packages for TA's but other local config could be added in the same model. 

Would like to merge this to keep the upstream updated.


pabelanger:
I'd like to see some spec tests here, ensuring there is code coverage and things work as expected.

manifests/ta.pp
```diff
@@ -0,0 +1,70 @@
...
+#   inputs          => {
+#     'monitor:///var/log/messages' => {
+#       'attributes' => {
+#         'index' => 'server_t',
+#       }
+#     }
+#   }
+# }
+#
+define splunk::ta (
```
hunner: Would this be easier understood if it was named `splunk::addon` or something? "ta" isn't something I would know without reading docs.

manifests/ta.pp
```diff
@@ -0,0 +1,70 @@
...
+  $splunk_home = '/opt/splunkforwarder',
+  $install_package = true,
```
hunner: We usually call parameters like this `package_manage`

manifests/ta.pp
hunner: Or `splunk::app` since it goes in the apps directory?

manifests/ta.pp
```diff
@@ -0,0 +1,70 @@
...
+  file { "${splunk_home}/etc/apps/${name}/local":
+    ensure => directory,
+  }
+
+  concat { "splunk::ta::inputs_${name}":
+    path    => "${splunk_home}/etc/apps/${name}/local/inputs.conf",
+    require => File["${splunk_home}/etc/apps/${name}/local"]
+  }
+
+   ## Stop concat::fragment from failing if no inputs are defined.
```
hunner: Can this be conditionalized to handle the case where there are no inputs defined? Having a blank fragment is odd.

Concat 2 will create an empty file even if there are no fragments.

manifests/ta.pp
hunner: Actually, https://answers.splunk.com/answers/477/what-is-the-difference-between-an-add-on-and-an-app.html

manifests/ta/input.pp
```diff
@@ -0,0 +1,13 @@
+define splunk::ta::input (
```
hunner: If this is a private define, please comment as such in the header. And maybe use `assert_private()` from stdlib

manifests/ta/input.pp
```diff
@@ -0,0 +1,13 @@
...
+  $ta,
```
hunner: This parameter is not self-explanitory. Is there a better name for this?

templates/ta/_input.erb
```diff
@@ -0,0 +1,8 @@
+[<%= @name %>]
+<%
+@attributes.each do |key,val|
+-%>
```
hunner: It's a silly change, but could you lump the `<%` and ruby code and `-%>` all on one line?

hunner:
Thanks for this! I think we don't require tests because there are none for the module at all, but if you could address my review comments that would be great!

crayfishx:
@hunner thanks for looking this PR over... I agree with your points and I've modified the PR with a second commit that covers these.

PR #28, opened 2016-02-03
Fix dependencies

mpdude:
Not sure if this is redundant or if things don't work for me because I don't have matching `File` resources


mpdude:
Without this, something is wrong for me when installing the forwarder: Some variable is undefined and the `package` points to a path that contains an `undef` value.

PR #21, opened 2015-08-04
Fix errors for forwarder.pp using rpm package provider

barrypitman:
I tried to install splunk::forwarder using the default package provider (rpm), but I ran into errors. This PR resolves those issue for me on a new RH7 machine.

PR #29, opened 2016-02-25
forwarder::pkg_provider should default to $splunk::params::pkg_provider

hunner:
Closes #21

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -54,7 +54,7 @@ class splunk::forwarder (
   $splunkd_listen    = '127.0.0.1',
   $purge_inputs      = false,
   $purge_outputs     = false,
-  $pkg_provider      = undef,
+  $pkg_provider      = $splunk::params::pkg_provider,
   $forwarder_confdir = $splunk::params::forwarder_confdir,
   $forwarder_output  = $splunk::params::forwarder_output,
   $forwarder_input   = $splunk::params::forwarder_input,


hunner:
This is backwards-incompatible

hunner:
The base class defaults to using `$splunk::params::pkg_provider` as the chosen package provider but the `splunk::forwarder` define defaults to undef for the provider. This commit now syncronizes the behavior to be consistent between both items.

hunner:
@asquelt @kyledecot Does this look appropriate to you? Would this be considered a "bug fix" because the behavior of the module was inconsistent? (re #21)

manifests/forwarder.pp
```diff
@@ -54,7 +54,7 @@
   $splunkd_listen    = '127.0.0.1',
   $purge_inputs      = false,
   $purge_outputs     = false,
-  $pkg_provider      = undef,
+  $pkg_provider      = $splunk::pkg_provider,
```
Ziaunys: This should be $splunk::params::pkg_provider, right?

discreet:
I've had to change the code to implement this fix. Will this be updated and merged soon? Seems like it just needs a very simple minor change to the PR.

hunner:
Updated, and calling this a bugfix instead of backwards-incompatible feature as the intent (as shown in the base class) is to follow `splunk::params::pkg_provider` and this failed to follow the original intent.

PR #36, opened 2016-03-16
Add conditional logic for using the staging module

Ziaunys:
- If the provider is yum or apt we don't need to use the staging module
  and this logic exists in splunk::forwarder.
- Removes an unnecessary include of the staging module.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -57,17 +57,22 @@ class splunk (
   include staging
 
   $virtual_service = $splunk::params::server_service
-  $staged_package  = staging_parse($package_source)
   $staging_subdir  = $splunk::params::staging_subdir
 
   $path_delimiter  = $splunk::params::path_delimiter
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -54,20 +54,24 @@
...
 
   $path_delimiter  = $splunk::params::path_delimiter
-  $pkg_path_parts  = [$staging::path, $staging_subdir, $staged_package]
-  $pkg_source      = join($pkg_path_parts, $path_delimiter)
 
-  staging::file { $staged_package:
-    source => $package_source,
-    subdir => $staging_subdir,
-    before => Package[$package_name],
+  if $pkg_provider != undef and $pkg_provider != 'yum' and  $pkg_provider != 'apt' {
```
hunner: Should we just go ahead and change this to `if $pkg_provider == 'rpm' or $pkg_provider == 'deb'` ? I wonder if anyone uses this on windows too, since it has some windows stuff in the module. Perhaps the msi provider?

PR #37, opened 2016-03-16
Update the resource ordering for the new Splunk types.

Ziaunys:
- Ensures that all the Splunk resource types are applied in the correct order.
- Depends on #34 and #35 
- I've only tested this on a CentOS 7 node.


bmjen:
:+1:

PR #38, opened 2016-03-17
Various updates and fixes

Ziaunys:
- Adds the ability to run Splunk as a non-root user on posix systems.
- Adds the splunkforwarder_web type and provider from Nick Perry's fork.
- Adds purging for all the new Splunk types.
- Provisions all the config files to ensure they exist beforehand.
- Properly tags virtual resources.
- Corrects some permission issues.
- Depends on #37 and this will need a rebase once its merged.

### DIFF
--- a/lib/puppet/provider/splunkforwarder_web/ini_setting.rb
+++ b/lib/puppet/provider/splunkforwarder_web/ini_setting.rb
@@ -1,0 +1,17 @@
+Puppet::Type.type(:splunkforwarder_web).provide(
+  :ini_setting,
+  # set ini_setting as the parent provider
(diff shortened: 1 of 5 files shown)


bmjen:
@Ziaunys how come you needed to abstract the ini_setting into a custom type/provider? It seems like an unnecessary abstraction, especially since the platform specific path for `forwarder_confdir` are set up in params.pp and overridable in the `forwarder` class. Changing this behavior seems like it would make custom confdirs disallowed.

Ziaunys:
@bmjen I went and updated the splunkforwarder_web type to use the splunk_config meta resource. I'll eventually update more types to use this resource.

PR #40, opened 2016-03-29
Change file mode to string.

Ziaunys:
- Change the file mode parameter in the forwarder.pp/init.pp files to
  use a string instead of an integer.
- Fixes some issues in the README. I forgot to remove parts of the
  template.
- Without this change newer versions of Puppet will fail.


mattkirby:
+1

kfjohnson:
:ok_hand:

PR #41, opened 2016-03-29
Add acceptance test skeleton.

Ziaunys:
- Adds the acceptance skeleton from the puppet-module-skeleton.
- Adds initial class_spec tests for Splunk and Splunkforwarder classes.
- Includes the appropriate module dependencies for testing.
- This is just the foundation for adding more tests.

PR #42, opened 2016-04-01
Fix module errors discovered while testing.

Ziaunys:
- Depends on #41
- The **$splunk_user** variable was referenced but undeclared.
  for this repo.
- Update the build and version to correctly reference the most recent
  values.
- The **$staged_package** variable was referenced before declaration.
- Creates separate spec files for testing the **splunk** and
  **splunk::forwarder** classes.
- Removes the path parameter from the **enable_splunk** and 
  **enable_splunkforwarder** execs because it causes issues on platforms
  that use the systemd service provider. The resulting init script that
  the exec generates will not get loaded by the systemd-sysv-generate
  module.
- Fixes an issue where the **splunkweb** Service would change from 'stopped'
  to 'running' on every Puppet run. The pattern for discovering the
  running service had an incorrect regex.


kfjohnson:
+1

bmjen:
Great work @Ziaunys !

PR #43, opened 2016-04-01
Update README with type documentation.

Ziaunys:
- Adds basic documentation for the Class parameters and various types
  that the module provides.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -127,55 +127,183 @@ This virtual resource will get collected by the `::splunk::forwarder` class if i
 
 ## Reference
 
-###::splunk::params Parameters
+### Types
+
+
+* `splunk_config`: This is a meta resource used to configur defaults for all the splunkforwarder and splunk types.
+
+* `splunk_authentication`: Used to manage ini settings in [authentication.conf][authentication.conf-docs]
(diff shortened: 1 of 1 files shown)


kfjohnson:
+1

discreet:
@Ziaunys since you added the documentation i was curious if there is a way to have multiple settings/values for the same section when using the forwarder_input and forwarder_output types.

Ziaunys:
@discreet You could pass multiple forwarder inputs and outputs to the forwarder_input and forwarder_output parameters since it's just taking those hashes and calling create resources on them. I can update the docs to make that more clear.

discreet:
@Ziaunys that would be awesome. To clarify though in the params.pp we see this as the default to forwarder_output:

 $forwarder_output = {
    "tcpout_defaultgroup"          => {
      section                      => 'default',
      setting                      => 'defaultGroup',
      value                        => "${server}_${logging_port}",
      tag                          => 'splunk_forwarder'
    },

I'm looking to achieve an output.conf like below (*ip addresses removed):

[default]
defaultGroup = SplunkCluster
disabled = false
[tcpout:SplunkClusterProd]
server = $IPone:$port, $IPtwo:$port
useACK = true
maxQueueSize = 7MB

PR #46, opened 2016-05-04
Adding new default_host parameter

sharumpe:
I know you're not really set up for submissions at this point, but just in case this is helpful, here's a small change I've made to a fork that I'd love to see in the base.  :) 

The 'norm' at our location is to have syslog use the hostname instead of the fqdn.  Setting the sourcetype to syslog will result in those logs having the "correct" hostname, while others may not be parsed and will end up with the fqdn, which makes for more troublesome searches for that host.

TL;DR: setting the default_host rather than using the parsed value means that confusion can be resolved.

I have tested this on CentOS 7.2 and Ubuntu 14.04.

Thanks!

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,6 +50,7 @@ class splunk (
   $splunkd_port         = $splunk::params::splunkd_port,
   $splunk_user          = $splunk::params::splunk_user,
   $pkg_provider         = $splunk::params::pkg_provider,
+  $default_host         = $splunk::params::default_host,
   $splunkd_listen       = '127.0.0.1',
   $web_port             = '8000',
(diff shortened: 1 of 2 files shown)


sharumpe:
OK, I have just made another commit (to add support for the Suse osfamily) which is a very small commit but I didn't realize it would be rolled up with this.

bastelfreak:
Hi @sharumpe, are you still interested in this PR? If so, please rebase against our latest master.

dhoppe:
@sharumpe Ping

bastelfreak:
Hello. I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested in it.

PR #48, opened 2016-06-28
split out forwarder and server into separate classes

yodermk:
Hello,

We have a very weird issue with this class.  We are on Linux and are ONLY trying to install the Splunk forwarder, not the main server.  It turns out, however, that it has been trying to also instantiate the main server classes.  We were getting an error on the license_splunk exec that the splunk command was not found.

@gregswift and I have been banging our heads on this for a couple days, and found out that it works if the two execs for the main Splunk server were removed from the path.

This would seem to indicate that somehow it is realizing wildcard services in a way that is inconsistent with how it is being called.  We scrutinized our code and the code in this module, and can not find any code path that should be realizing these services.

We see this issue on Puppet 3.  On our Puppet 2 standup (which is on its way out), it works as expected.  We have not tried Puppet 4.

Obviously we're open to a better way to do this.  Or it may be a bug in this module or in Puppet itself.  We're just submitting this PR to give a heads up on the problem and to see if anyone else has hit it or has another idea for a solution.

Thank you!

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -118,8 +118,8 @@ class splunk::forwarder (
   # there is non-generic configuration that needs to be declared in addition
   # to the agnostic resources declared here.
   case $::kernel {
-    'Linux': { class { 'splunk::platform::posix': splunkd_port => $splunkd_port,
-                                                  splunk_user  => $splunk_user } }
+    'Linux': { class { 'splunk::platform::posix::forwarder': splunkd_port => $splunkd_port,
+                                                             splunk_user  => $splunk_user } }
     'SunOS': { include splunk::platform::solaris }
     default: { } # no special configuration needed
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -173,8 +173,8 @@ class splunk (
   # there is non-generic configuration that needs to be declared in addition
   # to the agnostic resources declared here.
   case $::kernel {
-    'Linux': { include splunk::platform::posix   }
-    'SunOS': { include splunk::platform::solaris }
+    'Linux': { include splunk::platform::posix::server   }
+    'SunOS': { include splunk::platform::solaris         }
     default: { } # no special configuration needed
   }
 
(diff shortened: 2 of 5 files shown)


gregswift:
The issue we were seeing was:

```
Info: Applying configuration version '1467137689'
Error: Could not find command 'splunk'
Error: /Stage[main]/Splunk::Platform::Posix/Exec[license_splunk]/returns: change from notrun to 0 failed: Could not find command 'splunk'
Notice: /Stage[main]/Splunk::Platform::Posix/Exec[enable_splunk]: Dependency Exec[license_splunk] has failures: true
Warning: /Stage[main]/Splunk::Platform::Posix/Exec[enable_splunk]: Skipping because of failed dependencies
Notice: /Stage[main]/Splunk::Virtual/Service[splunk]: Dependency Exec[license_splunk] has failures: true
Warning: /Stage[main]/Splunk::Virtual/Service[splunk]: Skipping because of failed dependencies
```

We don't load the main splunk class... we were able to reproduce with just this block:

```
 # setup splunk params
    class { '::splunk::params':
      server   => $splunk_server,
    }

    # install splunk forwarder
    class { '::splunk::forwarder':
      pkg_provider => 'yum',
    }
```

I extracted the license_splunk relationships from the expanded_relationships.dot graph on puppet 2 and puppet3.
license_splunk_p2_exp_rel
license_splunk_p3_exp_rel


We are running Puppet 3.8.1.  (thought we were on 3.8.7 but apparently thats just clients). One of our peer teams is on 3.8.5 and is not seeing the issue. We are going to update and see what happens.

yodermk:
Just updated to 3.8.7, same error.

bastelfreak:
Hi @yodermk, are you still interested in this PR? If so, can you take a look at the merge conflicts and rebase?

yodermk:
Crud, I didn't see this. Someone else attempted to update to upstream and did apparently this this problem.
I'm on a different team now, and we're working on standing up Puppet 4. We will try the upstream module with that. If it doesn't work I will revisit this.  Thanks!

PR #51, opened 2016-09-26
Modulesync 0.12.7

bastelfreak:


.sync.yml
```diff
@@ -0,0 +1,4 @@
+---
+.travis.yml:
+	secure: "RrAmWtM6eTjZZzD954AIgR179Pqp14lzHhd/C9cpKbTPynLncuim08CEvjmq+7pgAy9XDg1d02x3udfZt4btR1sBdyNRpNN3yUhWptmGU61HRJdiZq+nSLQkIYsqXanhk+O3NndFojO58WRD01dkWEc6HRHjlivuYNxDXmMkg9k="
```
alexjfisher: Sacre bleu!  A 'tab'?

manifests/password.pp
```diff
@@ -40,18 +40,18 @@
...
-    content => $password_content,
     ensure  => present,
+    content => $password_content,
     require => Package[$package_name],
     notify  => Service[$virtual_service],
     tag     => "splunk_password",
   }
 
   file { "$secret_file":
+    ensure  => present,
```
alexjfisher: Somehow, I thought this was a lint error in itself and we preferred `ensure => file`?

alexjfisher: I guess we don't use https://github.com/voxpupuli/puppet-lint-file_ensure-check?

.sync.yml
bastelfreak: shitty vim paste mode :(

manifests/password.pp
bastelfreak: jup

PR #53, opened 2016-09-26
Bug fixes, lint, rubocop, spec tests etc.

alexjfisher:
A couple of bug fixes.  Lots of lint fixes.  Tests (hardly any) now at least run.  See individual commits for details.


lib/puppet/provider/ini_setting/splunk.rb
```diff
@@ -13,7 +13,7 @@ def self.file_path
     File.join(@file_path, file_name)
   end
 
-  def self.set_file_path(path)
+  def self.set_file_path(path) # rubocop:disable Style/AccessorMethodName
```
alexjfisher: I wimped out here.  Somebody with more knowledge of the module and how the inifile providers work could have a go if they really wanted to.  Probably not worth it though.

alexjfisher:
@crayfishx Would you like to review this one?

lib/puppet/provider/ini_setting/splunk.rb
crayfishx: According to the rubocop docs we should rename this to 

```ruby
def self.file_path=(path)
```

And then change the references to this method in splunk_config

https://github.com/voxpupuli/puppet-splunk/blob/master/lib/puppet/type/splunk_config.rb#L95-L104

eg;

```
Puppet::Type.type(res_type).provider(:ini_setting).file_path=(self[:server_confdir])
```

alexjfisher: > 
```ruby
def self.file_path=(path)
```

leads to `Style/TrivialAccessors: Use attr_writer to define trivial writer methods.` which can be fixed by replacing the whole thing with 
```ruby
attr_writer :file_path
```
I think?

>
```ruby
Puppet::Type.type(res_type).provider(:ini_setting).file_path=(self[:server_confdir])
```
leads to `Style/SpaceAroundOperators: Surrounding space missing for operator =` at which point it was late and I got scared.

crayfishx: I think you'll need....

```ruby
class << self
  attr_writer :file_path
end
```

PR #55, opened 2016-09-29
Added rspec tests for types and splunk_config and other bugfixes

crayfishx:
* Added rspec tests for splunk types
* Added rspec tests for splunk_config
* Bugfix: splunk_input file_path not being set
* Bugfix: calling provider file_path method from a variable

### DIFF
--- a/lib/puppet/type/splunk_config.rb
+++ b/lib/puppet/type/splunk_config.rb
@@ -90,8 +90,8 @@ Puppet::Type.newtype(:splunk_config) do
       :splunk_transforms,
       :splunk_web
     ].each do |res_type|
-      klass = Puppet::Type.type(res_type).provider(:ini_setting)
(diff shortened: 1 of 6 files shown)


alexjfisher:
@crayfishx Did https://github.com/voxpupuli/puppet-splunk/pull/53/commits/bead52a8543589584c3ff5f862c13874965653ce break the module?

crayfishx:
@alexjfisher Im not entirely sure - but calling  `Puppet::Type.type(:foo).provider(:ini_file).file_path =` didn't work when I was writing the rspec tests, but does work when it is assigned to a variable - which is a bit odd - either way it reads better and is more understandable as a variable assignment to make it clear that we're calling a class method in the provider.

apenney:
@crayfishx I'd prefer a nice Fira Mono, possibly italicized..

PR #66, opened 2016-10-31
Add missing badges

dhoppe:


README.md
```diff
@@ -35,8 +40,10 @@ or apt to install these components if they're self-hosted.
 
 ### What splunk affects
 
-* Installs the Splunk/Forwarder package and manages their config files. It does not purge them by default.
-* The module will set up both Splunk and Splunkforwarder to run as the 'root' user on POSIX platforms.
+* Installs the Splunk/Forwarder package and manages their config files. It does
+  not purge them by default.
+* The module will set up both Splunk and Splunkforwarder to run as the 'root'
+  user on POSIX platforms.
```
crayfishx: Is there a reason for wrapping lines in the README.md?  I didn't realize that was "a thing"

?

crayfishx: My personal preference would be to disable this check - maintaining documentation with line breaks is just a PITA - not sure what the wider preference is though.

dhoppe: I disagree, because I changed this for almost every module and this is more readable in text editors like VIM.

@bastelfreak What do you think?

bastelfreak: IMO we should honour the linter here. My vim makes automatic linebreaks after X chars, that needs to be configured but works fine.

@crayfishx which parts do you recognize as PITA?

crayfishx: Adding three words to a sentence means I then need to manually adjust all the line spacing for the rest of the paragraphs.... yeah I know there are editor plugins that can do this, I just don't see the point in enforcing this - I quite like text to wrap to the length I choose by resizing my terminal window.... maybe it's just me ;-)

PR #73, opened 2016-11-09
Make the file reference use the variable and not a fixed path

ralfbosz:
Find it curious that the "file" references in forwarder.pp are not using the variable from the params.pp, small fix. Got the errror when trying to deploy on Windows, but guess it would also occur when using another forwarder_confdir...

For Windows support it seems to me support needs to be added to install the universalforwarder as a chocolatey package, since I don't see how it is gonna work using the msi only...

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -159,17 +159,17 @@ class splunk::forwarder (
     mode => '0644',
   }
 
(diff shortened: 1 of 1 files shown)


crayfishx:
LGTM - makes sense.

PR #87, opened 2017-02-08
Allow redirecting of settings into custom contexts

michaelweiser:
Add a new resource parameter that allows redirection of settings into
contexts other than system/local. In prarticular this allows to
customize (and even create) apps with context =>
"apps/<app>/{default,local}". Incidentally this centralises redundant
occurences of system/local as the parameter default.

### DIFF
--- a/lib/puppet/provider/ini_setting/splunk.rb
+++ b/lib/puppet/provider/ini_setting/splunk.rb
@@ -14,6 +14,10 @@ Puppet::Type.type(:ini_setting).provide(
   def self.file_path
     raise Puppet::Error, 'file_path must be set with splunk_config type before provider can be used' if @file_path.nil?
     raise Puppet::Error, 'Child provider class does not support a file_name method' unless respond_to?(:file_name)
-    File.join(@file_path, file_name)
+    @file_path
+  end
+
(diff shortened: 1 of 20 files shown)


michaelweiser:
Phew. Sorry to keep repushing changed commits multiple times but the test suite kinda got to me. It seems to be sorted now.

bastelfreak:
@crayfishx could you have a look please?

bastelfreak:
Thanks for the PR @michaelweiser!

lib/puppet/provider/ini_setting/splunk.rb
```diff
@@ -14,6 +14,10 @@ class << self
   def self.file_path
     raise Puppet::Error, 'file_path must be set with splunk_config type before provider can be used' if @file_path.nil?
     raise Puppet::Error, 'Child provider class does not support a file_name method' unless respond_to?(:file_name)
-    File.join(@file_path, file_name)
+    @file_path
```
alexjfisher: This breaks `instances` (as the parent provider `instances` method no longer uses the correct path).  In turn this breaks all settings related to purging unmanaged resources.

PR #89, opened 2017-02-09
Ensure /etc/init.d/splunk is created before splunk services

treydock:


iamjamestl:
Thank you.  This works for me on RHEL 7 (systemd).

PR #90, opened 2017-02-09
Allow splunk server services to be overridden

treydock:
Really useful to set `server_service => 'splunk'` since all the other services are started/stopped by the splunk init.d script.


roidelapluie:
Thanks!

PR #91, opened 2017-02-09
Fix resource ordering in splunk class so that splunk types will properly notify services

treydock:


bastelfreak:
Hi @treydock. thanks for the PR!

Issue #69, opened 2016-11-01
The splunk::forwarder::purge_inputs generates error 


Vincent--:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.7
- Ruby: ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
- Distribution: Linux siem 4.4.0-31-generic #50~14.04.1-Ubuntu SMP Wed Jul 13 01:07:32 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
- Module version: 5.0.1

### How to reproduce (e.g Puppet code you use)
Try to enable splunk::forwarder::purge_inputs

```
splunk::forwarder::purge_inputs     : true
```

### What are you seeing

An error message is generated during the provision:
```
Notice: Compiled catalog for puppet in environment production in 4.92 seconds
Error: /Stage[main]/Splunk::Params/Splunk_config[splunk]: Failed to generate additional resources using 'generate': can't convert nil into String
Notice: Finished catalog run in 15.65 seconds
```

### What behaviour did you expect instead
No error message

callafeucht:
I am also encountering this error.

- Puppet: 3.7.5
- Ruby: ruby 2.0.0p598 (2014-11-13) [x86_64-linux]
- Distribution: CentOS Linux release 7.2.1511 (Core) 
- Module version: v5.0.2

michaelburgener:
I'm having the same issue as well

- Puppet: 4.8.1
- Ruby: ruby 2.1.9p490 (2016-03-30 revision 54437) [x86_64-linux]
- Distribution: Ubuntu 16.04.1 LTS Linux 4.4.0-51-generic x86_64
- Module version: v5.1.1-rc0

wottah:
+1
Also encountered this problem. The issue seems to come from within the inifile dependency.

TraGicCode:
This was fixed in #96 and i also tested this locally with beaker.

PR #96, opened 2017-02-28
types: Fix purging when section contains '//'

iamjamestl:
A change made in 0fe96f7 to support section headings of the form:

```
foo://bar
```

broke purging, where the underlying inifile type generates resources
with titles like:

```
section/setting
```

When a resource is generated with a title 'foo://bar/setting', the
setting name is not parsed.  This is the case with inputs which often
take on the form:

```
[monitor:///var/log/messages]
setting1 = value1
setting2 = value2
```

The result is an error message: "Error:
/Stage[main]/Splunk::Params/Splunk_config[splunk]: Failed to generate
additional resources using 'generate': Got nil value for setting."  See:
issue #69.

This change looks for additional slashes beyond the '://' and will
parse out the following string as the setting name.

This is not very satisfying because it could break a theoretical section
name containing extra slashes, but I have not come across that before,
and the `[monitor:///../../../]` case is **very** common.

### DIFF
--- a/lib/puppet_x/puppetlabs/splunk/type.rb
+++ b/lib/puppet_x/puppetlabs/splunk/type.rb
@@ -7,9 +7,14 @@ module PuppetX
 
           type.define_singleton_method(:title_patterns) do
             [
-              [%r{^([^\/]*)$},   [[:section]]],
-              [%r{^(.*\/\/.*)$}, [[:section]]],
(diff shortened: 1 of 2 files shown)


roidelapluie:
Travis did not run.

iamjamestl:
Fixed--just minor rubocop violations.

roidelapluie:
Thanks!

PR #99, opened 2017-03-13
Set `forwarder_install_options` to be `undef` for OS other than Windows.

shadow999:


bastelfreak:
Hi @shadow999, thanks for the PR!

Issue #100, opened 2017-03-24
Windows Installs Will Never Work with Default Options


kenmaglio:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:  4x
- Ruby: 
- Distribution: PE
- Module version: Latest

## How to reproduce (e.g Puppet code you use)
Deploy splunk::forwarder to windows with default options

## What are you seeing
on lines in params
These default paths will NEVER let the MSI actually install using the package manager.
as these are directly included into install arguments. please see this note on puppet docs on Package resource:

With Windows packages, note that file paths in an install option must use backslashes. (Since install options are passed directly to the installation command, forward slashes won’t be automatically converted like they are in file resources.) Note also that backslashes in double-quoted strings must be escaped and backslashes in single-quoted strings can be escaped.

  if $::osfamily == 'Windows' {
    $forwarder_dir = pick($forwarder_installdir, 'C:/Program Files/SplunkUniversalForwarder')
    $server_dir    = pick($server_installdir, 'C:/Program Files/Splunk')
  } else {
    $forwarder_dir = pick($forwarder_installdir, '/opt/splunkforwarder')
    $server_dir    = pick($server_installdir, '/opt/splunk')
  }

TraGicCode:
I'm having the same issue along with some other issues on windows.

PR #102, opened 2017-04-12
Fix Windows Forwarder Installdir

TraGicCode:
The Installdir for windows must contain backslashes.  The default values never worked on windows.  Relates to #100


TraGicCode:
closes #100

PR #107, opened 2017-04-18
This commit sets the splunk_user for Windows to 'administrator'

ralfbosz:
There was only one 'default' which was 'root', but this user is
non-existent on Windows.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -90,9 +90,11 @@ class splunk::params (
   if $::osfamily == 'Windows' {
     $forwarder_dir = pick($forwarder_installdir, 'C:\\Program Files\\SplunkUniversalForwarder')
     $server_dir    = pick($server_installdir, 'C:/Program Files/Splunk')
+    $splunk_user   = 'Administrator'
   } else {
     $forwarder_dir = pick($forwarder_installdir, '/opt/splunkforwarder')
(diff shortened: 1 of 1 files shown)


bastelfreak:
HI @ralfbosz, thanks for the commit. Please take a look at the used email address in your commit. It isn't asociated with your github account.

ralfbosz:
fixed it...

vinzent:
@bastelfreak @ralfbosz is Administrator the correct value for this?

according the README the splunk_user param is *the user to run Splunk as.*. On Windows sytstems this isn't Administrator normally. I think the default is LocalSystem or something like this.

splunk_user will also set the ownership of some files to Administrator. I don't know if this is correct.

ralfbosz:
The LocalSystem account is a predefined local account used by the service control manager. This account is not recognized by the security subsystem, so you cannot specify its name in a call to the LookupAccountName function. 

https://msdn.microsoft.com/nl-nl/library/windows/desktop/ms684190(v=vs.85).aspx

On a Windows 2012 R2-server:

C:\windows\system32>puppet resource user localsystem
user { 'localsystem':
  ensure => 'absent',
}

Seems to me that user isn't usable in Puppet. A better solution would be making the user 'configurable'.

PR #108, opened 2017-05-03
This commit adds the pkg_provider 'chocolatey' to the module

ralfbosz:
When using 'chocolatey' as pkg_provider it is not needed to
include :staging, this is excluded in the init.pp and forwarder.pp

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -72,7 +72,7 @@ class splunk::forwarder (
 
   $path_delimiter  = $splunk::params::path_delimiter
   #no need for staging the source if we have yum or apt
-  if $pkg_provider != undef and $pkg_provider != 'yum' and  $pkg_provider != 'apt' {
+  if $pkg_provider != undef and $pkg_provider != 'yum' and $pkg_provider != 'apt' and $pkg_provider != 'chocolatey' {
     include ::staging
 
     $src_pkg_filename = staging_parse($package_source)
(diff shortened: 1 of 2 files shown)


ralfbosz:
We're using Chocolatey as package-provider, for that we need to 'skip' the include of ::staging, this PR fixes that.

manifests/forwarder.pp
```diff
@@ -153,10 +157,17 @@
   Package[$package_name] -> Splunkforwarder_transforms<||> ~> Service[$virtual_service]
   Package[$package_name] -> Splunkforwarder_web<||>        ~> Service[$virtual_service]
 
-  File {
-    owner => $splunk_user,
-    group => $splunk_user,
-    mode => '0644',
+  if $::osfamily == 'windows' {
```
bastelfreak: Could you migrate this to the new facts hash? Probably $facts['os']['family'].

bastelfreak: this should also be rewritten to the selector pattern. An example is here: https://github.com/voxpupuli/puppet-splunk/pull/114/files. In my opinion it's a bit more beautiful to write the result of the selector into a variable, and use that in the file resource declaration.

----

This should be rewritten to only declare the file resource once

bastelfreak:
Hi @ralfbosz, sorry for the late feedback. Could you please rebase against latest master and take a look at the inline comments?

manifests/forwarder.pp
ralfbosz: see that this has already been merged, will update my PR...

ralfbosz:
When pkg_provider is "chocolatey" the source needs to be 'empty', so added selector, but being strict it's a selector in a selector with the 'pick' which is also there...

PR #113, opened 2017-05-26
Allow for installing packages from UNC paths on Windows

natemccurdy:
This PR incorporates commits that ultimately allow Windows agents to install packages from UNC paths.

Prior to this, if setting `$package_source` to a UNC path on Windows, the `package` resource wouldn't ever use the value of `$package_source` as it's source. The reason for that is because a UNC path doesn't need to be staged with `staging::file` thus the `$pkg_source` variable was never set.

While I was fixing that problem, I also cleaned up the variable names in the `staging::file` steps to make their intent more clear.


bastelfreak:
Thanks!

PR #119, opened 2017-07-25
Fixing global varaible reference

TraGicCode:
It looks like whoever added the server_service parameter to the posix class left of the ::params::.  This is Preventing others from writing unit tests against this.

@treydock can you verify this was a mistake/typo?


treydock:
@TraGicCode Not using params was intentional, so that splunk class could redefine the service list.  This was done similar to splunkd_port.

treydock:
@TraGicCode This issue is likely because the `splunk::server_service` variable isn't defined when dealing only with `splunk::forwarder` class.

TraGicCode:
@treydock I'll take a look at this later! thanks for the feedback

PR #126, opened 2017-09-01
wire up splunkforwarder_server type

alexcreek:
The splunkforwarder_server type was already created but it wasn't added to splunk_config.rb.  Added.


TraGicCode:
LGTM 👍

PR #128, opened 2017-09-13
Replace staging module with archive module

### LINKED ISSUE #127: Replace Staging with archive module
Lets get the staging module, which is deprecated, replaced with the archive module like was done with the rabbitmq module!

TraGicCode:
This will get rid of the staging module being used and replace it with the newer recommended archive module.

Closes #127 

@wyardley Can you review this?


TraGicCode:
I Still need to test this on windows and other nodes besides ubuntu.  Just need a set of eyes to look at this.

wyardley:
Can you suppress the merge (rebasing is better), and squash it down into either 1 commit or a few commits that are broken up into logical functions. 

Did you run the acceptance tests yet?

manifests/params.pp
```diff
@@ -80,7 +80,7 @@
   # Based on the small number of inputs above, we can construct sane defaults
   # for pretty much everything else.
 
-  # Settings common to everything
+  # # Settings common to everything
```
wyardley: Is the extra comment mark a mistake here?

TraGicCode: Yes.  Was when i was commenting out staging to get everything to fail and then replace it

TraGicCode: Done

manifests/platform/solaris.pp
```diff
@@ -13,11 +13,11 @@
 # Requires: nothing
 #
 class splunk::platform::solaris inherits splunk::virtual {
-  include ::staging
+  include ::archive
   include ::splunk::params
   include ::splunk::platform::posix
 
-  $path         = $staging::path
+  $path         = $archive::path
```
alexjfisher: There's no such variable as $archive::path

spec/spec_helper_system.rb
```diff
@@ -19,7 +19,7 @@
 
     # Install modules and dependencies
     puppet_module_install(source: proj_root, module_name: 'splunk')
-    shell('puppet module install nanliu-staging --version 0.3.1')
+    shell('puppet module install puppet-archive --version 0.3.1')
```
alexjfisher: There's no such version of puppet/archive.

TraGicCode: Done.

TraGicCode: Hey @alexjfisher It does work but i might be accessing this the wrong way from the puppet archive module

image

manifests/platform/solaris.pp
TraGicCode: Hey @alexjfisher It does work but i might be accessing this the wrong way from the puppet archive module

image

alexjfisher: My mistake.  I missed the `inherits` https://github.com/voxpupuli/puppet-archive/blob/master/manifests/init.pp#L28

alexjfisher: The staging module created $staging::path for you.  With archive, I think you'll need to `include ::archive::staging` instead of just `::archive` ??

TraGicCode: Done.

alexjfisher: Also needed in `forwarder.pp` ?

manifests/forwarder.pp
```diff
@@ -73,16 +73,16 @@
   $path_delimiter  = $splunk::params::path_delimiter
   #no need for staging the source if we have yum or apt
   if $pkg_provider != undef and $pkg_provider != 'yum' and $pkg_provider != 'apt' and $pkg_provider != 'chocolatey' {
-    include ::staging
+    include ::archive
```
alexjfisher: `include ::archive::staging` here too?

TraGicCode:
Before this can be merged there is one thing that **will break all existing windows users that don't have chocolatey**.  Briefly looking, by default, it looks like the archive class in the archive module assumes the user has 7zip installed and if not it will install it via chocolatey.  

Therefore i need to do the following:

- [ ]  See if puppet-archive will allow downloading a file without requiring 7zip and chocolatey

manifests/forwarder.pp
TraGicCode: Done.

alexjfisher:
If you're not needing to extract anything you download, I don't see why you'd need to `include ::archive`.  (You would then need to create and use your own 'staging' path though).

spec/spec_helper_system.rb
```diff
@@ -19,7 +19,7 @@
 
     # Install modules and dependencies
     puppet_module_install(source: proj_root, module_name: 'splunk')
-    shell('puppet module install nanliu-staging --version 0.3.1')
+    shell('puppet module install puppet-archive --version 2.0.0')
```
wyardley: btw, it seems like this is for https://github.com/puppetlabs/rspec-system, which is deprecated and presumably not in use now that the module has beaker based acceptance tests

manifests/platform/solaris.pp
TraGicCode: Done

TraGicCode:
I tested on windows and it looks like that block never executes that uses the archive module.  If you pass in a folder location where the msi exists it will use that existing file to install the forwarder.  If you pass in an HTTP/S url the package resource will stream the msi and perform the install without ever saving the file on the windows system.  @alexjfisher From what i can tell this should be good.  Tested on 

Ubuntu 16.04 ( server and forwarder )
Windows ( only forwarder..looks like server has never worked )

spec/spec_helper_system.rb
```diff
@@ -0,0 +1,27 @@
...
+  c.include RSpecSystemPuppet::Helpers
+
+  # This is where we 'setup' the nodes before running our tests
+  c.before :suite do
+    # Install puppet
+    puppet_install
+
+    # Install modules and dependencies
+    puppet_module_install(source: proj_root, module_name: 'splunk')
+    shell('puppet module install puppet-archive --version 2.0.0')
```
bastelfreak: please take a look at the spec_helper_acceptance here:
https://github.com/voxpupuli/puppet-grafana/pull/61/commits/d09835a9453c50ea4c803a15fe2f44f1e57542e3

install_module_dependencies installs all modules from the metadata.json, install_module_from_forge is for additional dependencies.

TraGicCode: After #136 is merged i can do this.

spec/spec_helper_system.rb
```diff
@@ -0,0 +1,27 @@
+require 'rspec-system/spec_helper'
```
ekohl: I believe ```spec_helper_system.rb``` is no longer used. Nowadays we only use ```spec_helper_acceptance.rb```

spec/spec_helper_system.rb
TraGicCode: @bastelfreak Done.

spec/spec_helper_system.rb
TraGicCode: I'll get that as well.

TraGicCode: Done.

spec/spec_helper_system.rb
TraGicCode: No Longer applicable

TraGicCode:
I believe i've cleared up all of the raised issues

bastelfreak:
@TraGicCode looks good to me. @alexjfisher can you check this PR again?

PR #129, opened 2017-09-13
In a previous issue i fixed the missing concat dependency but did not update the acceptance tests.

TraGicCode:


spec/spec_helper_acceptance.rb
```diff
@@ -20,6 +20,7 @@
       on host, puppet('module', 'install', 'puppetlabs-stdlib'), acceptable_exit_codes: [0, 1]
       on host, puppet('module', 'install', 'nanliu-staging'), acceptable_exit_codes: [0, 1]
       on host, puppet('module', 'install', 'puppetlabs-inifile'), acceptable_exit_codes: [0, 1]
+      on host, puppet('module', 'install', 'puppetlabs-concat'), acceptable_exit_codes: [0, 1]
```
wyardley: btw, switching to `beaker/module_install_helper` can have it auto-install the dependencies based on the metadata, in a way that's a bit more readable.

See, e.g.,
https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/spec_helper_acceptance.rb#L7-L13
https://github.com/voxpupuli/puppet-nodejs/blob/master/spec/spec_helper_acceptance.rb#L5-L13

PR #130, opened 2017-09-14
SplunkForwarder has no concept of a server_service. 

TraGicCode:
when setting up a splunk forwarder on a posix based system you get a warning because the the default parameter for the splunk::platform::posix class is set to $splunk::server_service which doesn't exist when you create a forwarder.  It doesn't exist because it doesn't make sense for a forwarder.  While there should be more cleanup this is a simple non-breaking change to fix the annoying warning.  A bigger refactor would be required to get this done correctly.

#119

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -146,7 +146,13 @@ class splunk (
   # there is non-generic configuration that needs to be declared in addition
   # to the agnostic resources declared here.
   case $::kernel {
-    'Linux': { include ::splunk::platform::posix   }
+    'Linux': {
+      class { '::splunk::platform::posix':
+        splunkd_port   => $splunkd_port,
+        splunk_user    => $splunk_user,
+        server_service => $server_service,
+      }
+    }
     'SunOS': { include ::splunk::platform::solaris }
     default: { } # no special configuration needed
   }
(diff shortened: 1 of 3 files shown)


wyardley:
I am not familiar with this module or Splunk, and also, can't even test it since the acceptance tests won't work even for the forwarder without a locally hosted package.

That said, would adding:
`server_service = undef` (or `server_service = []`)
in the line after
https://github.com/voxpupuli/puppet-splunk/blob/master/manifests/forwarder.pp#L140
work, and be a little simpler / less intrusive?

Maybe ask someone who has more experience with this module to review?

TraGicCode:
I agree with your recommendation.  This module needs to be cleaned up but that is a bigger beast alltogether.  ATM noone can run puppet-rspec against this because of it blows up saying hey there is a variable in this module that doesnt exist!

TraGicCode:
Done.

TraGicCode:
@wyardley I'm going to try and add a unit test for this to prevent regression and to prove out the fix.  Please wait before merging this.

TraGicCode:
The fix you recommended still has this issue.  Going to revert my revert ;)

image

TraGicCode:
My fix does indeed work.  It also reveals another variable with the same issue..  I will open another issue for this.


image

manifests/init.pp
```diff
@@ -146,7 +146,13 @@
   # there is non-generic configuration that needs to be declared in addition
   # to the agnostic resources declared here.
   case $::kernel {
```
bastelfreak: on a long term point of view, we should switch to $facts['kernel']

PR #134, opened 2017-09-20
SplunkForwarder has no concept of a splunkd_port

TraGicCode:


TraGicCode:
This fixes the other variable issue as well and also lets the compilation test pass for centos now for the forwarder :)! resolves #134

PR #139, opened 2017-09-24
6.2.0 Release

TraGicCode:
Cutting new release.  Did a bump on the minor since the public API didn't change and i simply updated a dependency.


TraGicCode:
There has to be a bump somewhere in the version in order to following semver.  Also I was basing the non-breaking change off the recommendation in the semver doc.

image

PR #144, opened 2017-09-25
Fix file_name for several providers

treydock:
This should fix new types/providers to work with this commit: https://github.com/voxpupuli/puppet-splunk/commit/54ece81e6c2938402e32c5f248262303bb2f470f

### DIFF
--- a/lib/puppet/provider/splunk_alert_actions/ini_setting.rb
+++ b/lib/puppet/provider/splunk_alert_actions/ini_setting.rb
@@ -3,6 +3,6 @@ Puppet::Type.type(:splunk_alert_actions).provide(
   parent: Puppet::Type.type(:ini_setting).provider(:splunk)
 ) do
   def self.file_name
-    'system/local/alert_actions.conf'
+    'alert_actions.conf'
   end
 end
--- a/lib/puppet/provider/splunk_deploymentclient/ini_setting.rb
+++ b/lib/puppet/provider/splunk_deploymentclient/ini_setting.rb
@@ -3,6 +3,6 @@ Puppet::Type.type(:splunk_deploymentclient).provide(
   parent: Puppet::Type.type(:ini_setting).provider(:splunk)
 ) do
   def self.file_name
-    'system/local/deploymentclient.conf'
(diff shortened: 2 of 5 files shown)


TraGicCode:
Hey @treydock .  Is the impact of this currently just now allowing the new types from working correctly?

treydock:
Without this changes the new types will write their configs to the wrong place.  I only ran in noop mode before downgrading to my fork, but looked like all the new types thought they had to be applied when they already existed.  So the new types are essentially broken without this change.

bastelfreak:
Can we add any type of unit/acceptance test for this?

TraGicCode:
@bastelfreak I added the unit tests for the types.  I tested locally they do indeed fail then made my changes to verify the bug is fixed.  Before merging this i want to ask @treydock if when i rebased his changes if i did one thing correctly.  It looks like the splunkforwarder_deploymentclient and deploymentclient types are identical to each other.  Is a mistake or intentional?

treydock:
@TraGicCode The duplication was intentional.  The reason is to be able to have one go to a Splunk server and one go to a forwarder, which is dictated by the tag.  The other reason is that the filesystem path for the forwarder's config is different.  The `file_path` for the forwarder types differs from non-forwarder types.

```
class profile::splunk::deployment_client inherits profile::params {

  $deployment_servers = query_nodes("Class[Profile::Splunk::Deployment_server] and puppet_environment=${::puppet_environment}", 'fqdn')
  $deployment_server  = $deployment_servers[0]

  $deploymentclient_configs = {
    'deployment-client/disabled' => 'false',
    'target-broker:deploymentServer/targetUri' => "${deployment_server}:8089",
  }

  $deploymentclient_configs.each |$name, $value| {
    @splunkforwarder_deploymentclient { $name: value => $value, tag => 'splunk_forwarder' }
    @splunk_deploymentclient { $name: value => $value, tag => 'splunk_server' }
  }

}
```

TraGicCode:
Thanks for clearing this up! @bastelfreak you can merge this whenever.

bastelfreak:
thanks for the contribution!

PR #145, opened 2017-09-26
Release v6.3.1

TraGicCode:
Doing a release so other people don't try and use the new types introduced last release and start creating issues.


bastelfreak:
merged + released

PR #150, opened 2017-10-01
[BREAKING CHANGE] Updated http download structure

TraGicCode:
This module used to be able to download splunk installer from splunk directly.  Unfortunately, when splunk updates their http url paths this module didnt get updated as well.  This updates the structure of where the installers should live along with hooking up docker containers for running in travis.  **Merging this would be considered a breaking change since the folder structure where people are placing the installers at on their local file system are changing to match how splunk NOW does it**.

Summary
- Updated expected folder structure for the splunk installers
- Updated Documentation of folder structure
- Updated default download location to be https://download.splunk.com in order to help people use this module easier instead of having to manually prestage the splunk installers.
- Updated acceptance tests to pull from download.splunk.com
- Updated .sync.yml to include the docker sets that should run under travis for beaker acceptance tests
- Fixed a bug where the forwarder is not actually idempotent because of a file mode. 🐛 
- In order to get acceptance testing working in docker i had to actually get this module to work while inside a docker container for the splunk server using OPTIMISTIC_ABOUT_FILE_LOCKING ( reference: https://github.com/chef-cookbooks/chef-splunk/blob/611be81ac8e533af7a71f0b2836e8de3aa346373/.kitchen.dokken.yml#L114-L126 )

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,6 +1,13 @@
 ---
 .travis.yml:
   secure: "RrAmWtM6eTjZZzD954AIgR179Pqp14lzHhd/C9cpKbTPynLncuim08CEvjmq+7pgAy9XDg1d02x3udfZt4btR1sBdyNRpNN3yUhWptmGU61HRJdiZq+nSLQkIYsqXanhk+O3NndFojO58WRD01dkWEc6HRHjlivuYNxDXmMkg9k="
+  docker_sets:
+  - set: docker/centos-6
+  - set: docker/centos-7
+  - set: docker/debian-7
+  - set: docker/debian-8
+  - set: docker/ubuntu-14.04
+  - set: docker/ubuntu-16.04
 spec/spec_helper.rb:
   spec_overrides: 
   - "require 'splunk_data.rb'"
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -103,6 +103,21 @@ class splunk (
     tag      => 'splunk_server',
   }
 
+  if $facts['virtual'] == 'docker' {
+    ini_setting { 'OPTIMISTIC_ABOUT_FILE_LOCKING':
+      ensure  => present,
+      section => '',
+      setting => 'OPTIMISTIC_ABOUT_FILE_LOCKING',
+      value   => '1',
+      path    => '/opt/splunk/etc/splunk-launch.conf',
+    }
+
+    Package[$package_name]
+    -> Ini_setting['OPTIMISTIC_ABOUT_FILE_LOCKING']
+    -> Exec <| tag   == 'splunk_server'  |>
+  }
+
+
   splunk_input { 'default_host':
     section => 'default',
     setting => 'host',
--- a/spec/acceptance/splunk_forwarder_spec.rb
+++ b/spec/acceptance/splunk_forwarder_spec.rb
@@ -5,7 +5,11 @@ describe 'splunk::forwarder class' do
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
       pp = <<-EOS
-      class { '::splunk::forwarder': }
+      class { '::splunk::params':
+      }
+      class { '::splunk::forwarder':
+        splunkd_port => 8090,
+      }
       EOS
 
       # Run it twice and test for idempotency
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,42 @@ script:
 matrix:
   fast_finish: true
   include:
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/centos-6 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/centos-7 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/debian-7 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/debian-8 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-14.04 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-16.04 CHECK=beaker
+    services: docker
+    sudo: required
   - rvm: 2.1.9
     bundler_args: --without system_tests development
     env: PUPPET_VERSION="~> 4.0" CHECK=test
--- a/README.md
+++ b/README.md
@@ -70,32 +70,40 @@ The files must be placed according to directory structure example given below.
 
 The expected directory structure is:
 
-      `-- files
-          |-- splunk
-          |   `-- $platform
-          |       `-- splunk-${version}-${build}-${additl}
-          `-- universalforwarder
-              `-- $platform
-                  `-- splunkforwarder-${version}-${build}-${additl}
+     $root_url/
+     └── products/
+         ├── universalforwarder/
+         │   └── releases/
+         |       └── $version/
+         |           └── $platform/
+         |               └── splunkforwarder-${version}-${build}-${additl}
+         └── splunk/
+             └── releases/
+                 └── $version/
+                     └── $platform/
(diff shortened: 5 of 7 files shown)


.sync.yml
```diff
@@ -1,6 +1,8 @@
 ---
 .travis.yml:
   secure: "RrAmWtM6eTjZZzD954AIgR179Pqp14lzHhd/C9cpKbTPynLncuim08CEvjmq+7pgAy9XDg1d02x3udfZt4btR1sBdyNRpNN3yUhWptmGU61HRJdiZq+nSLQkIYsqXanhk+O3NndFojO58WRD01dkWEc6HRHjlivuYNxDXmMkg9k="
+  docker_sets:
+  - set: docker/ubuntu-16.04
```
bastelfreak: Did you test if our centos/ubuntu14.04/debian8/debian9 (https://github.com/voxpupuli/modulesync_config/pull/388) work as well?

bastelfreak: Did you test if our centos/ubuntu14.04/debian8/debian9 (voxpupuli/modulesync_config#388) work as well?

spec/acceptance/splunk_forwarder_spec.rb
```diff
@@ -1,25 +1,31 @@
...
+# require 'spec_helper_acceptance'
 
-describe 'splunk::forwarder class' do
-  context 'default parameters' do
-    # Using puppet_apply as a helper
-    it 'works idempotently with no errors' do
-      pp = <<-EOS
-      class { '::splunk::forwarder': }
-      EOS
+# describe 'splunk::forwarder class' do
```
bastelfreak: isn't this spec file needed anymore, or did you disable it just for testing?

TraGicCode: Disabled just for testing

.sync.yml
TraGicCode: Only enabled this one because of testing

TraGicCode:
Hey guys I could use some help on this pr.  When trying to install splunk server with a a vagrant vm on ubuntu 16.04 everything works perfectly.  When I immediately try to use docker I get the following error as this guy posted

https://answers.splunk.com/answers/323671/splunk-light-running-on-docker-throwing-unusable-f.html

My question is where would be the best place to create this environment variable? I'm very new to docker.

.sync.yml
TraGicCode: Adding more of the docker sets now since i solved the issue with splunkserver running inside a docker container.

spec/acceptance/splunk_forwarder_spec.rb
TraGicCode: Reenabled these since tests are now working inside a docker container.

spec/acceptance/splunk_spec.rb
```diff
@@ -5,6 +5,11 @@
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
       pp = <<-EOS
+      class { '::splunk::params':
+        version => '7.0.0',
+        build   => 'c8a78efdd40f',
+        src_root => 'https://download.splunk.com',
```
wyardley: This should be unnecessary, no? (since it's the same as params)?

More generally, I'm not sure about passing stuff into params directly, though I know the module did this before as well.

spec/acceptance/splunk_forwarder_spec.rb
```diff
@@ -5,6 +5,12 @@
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
       pp = <<-EOS
+      class { '::splunk::params':
+        version      => '7.0.0',
+        build        => 'c8a78efdd40f',
+        src_root     => 'https://download.splunk.com',
+        splunkd_port => 8090,
```
wyardley: Ditto here, this should be mostly default (except port?)

manifests/init.pp
```diff
@@ -103,6 +103,21 @@
     tag      => 'splunk_server',
   }
 
+  if $facts['virtual'] == 'docker' {
+    ini_setting { 'OPTIMISTIC_ABOUT_FILE_LOCKING':
```
wyardley: Do you know if this is always needed in any Docker container? Or is it just required for the acceptance tests to work in Travis?

TraGicCode: Both.  I don't know the exact details but if you google *splunk docker unusable filesystem* This is pretty much the solution everyone provides.  Even the chef-splunk repo did this but this was for their acceptance test.  I Also could not find exact details on why/how this actually works.  If you want to try this out locally under docker remove that line and watch it fail with the above error message.

spec/acceptance/splunk_forwarder_spec.rb
TraGicCode: I had to do this because currently the acceptance test try to install both a splunk forwarder and a splunk enterprise instance on the same node.  Each product has a management utility that listens on port 8089 by default.  To get around this i just told the forwarder to be on the non-default port.

While this package does allow for uninstalling your splunk forwarder and splunk enterprise instance the uninstallation is technically not quite correct.  In order to prevent adding more scope to this already existing bug i did what i did above.

spec/acceptance/splunk_spec.rb
TraGicCode: I can remove this since i did change it to the defaults.  Shall i remove this?

Also yes this module uses params in a non-standard way.  Partly in fact because both the forwarder and the server try and reuse the params class.

manifests/init.pp
wyardley: Yeah, I saw the comment before. Just wondering whether it's best to do this in the module or in the acceptance test; I think a case could possibly be made either way, but seems kind of magical, esp. since there's no parameter to control it.

It sounds like if you setup the volumes correctly, it may not be necessary to set this?
https://hub.docker.com/r/alexanderm/docker-splunk/
```
To work around the unusable filesystem issue you need to mount in a volume from the host for the /opt/splunk/var/lib
For example I started docker like this:
docker run -v /data/splunk-test:/opt/splunk/var/lib -p 8000:8000 -it alexanderm/docker-splunk bash
```

spec/acceptance/splunk_forwarder_spec.rb
wyardley: But aren't version / build / src_root already set to those by default in params?
But if you're only setting the forwarder to a non-default port, can't `splunkd_port` be set in `splunk::forwarder` directly?

spec/acceptance/splunk_spec.rb
wyardley: I saw that postgresql module uses `postgresql::global` or something along those lines to work around setting things that are common to multiple classes (and then pulls those in / out of params as necessary) - not sure if that's a more standard pattern though... most modules these days seem to be leaning towards having the main class be the interface to everything, and then having toggles there (`enable_forwarder = true` in class `splunk`). Probably not worth changing in this PR, but might be worth asking about, and possibly changing down the road.

spec/acceptance/splunk_forwarder_spec.rb
TraGicCode: I suppose.  that is the problem this module makes it confusing with its use of params.pp

I will:
1.) remove the defaults from params.pp
2.) move the splunkd_port to the forwarder class

manifests/init.pp
TraGicCode: @wyardley I saw that but that isn't "Applicable" i believe in our case.  When running docker acceptance tests we are not taking a folder on the host and mapping it to a folder in the docker container.  Make sense?

spec/acceptance/splunk_forwarder_spec.rb
TraGicCode: Done.

spec/acceptance/splunk_spec.rb
TraGicCode: Yep I agree there are a couple different approaches.  That is a beast for another day!

manifests/init.pp
wyardley: Ah, yeah, I thought this change was in the manifest  itself, not in the test. I agree that this is a reasonable workaround in the acceptance test itself.

spec/acceptance/splunk_spec.rb
```diff
@@ -5,6 +5,8 @@
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
       pp = <<-EOS
+      class { '::splunk::params':
+      }
```
wyardley: I don't think you need to invoke `splunk::params` explicitly at all, do you?

TraGicCode: I agree :).  Let me see if I can remove this without things going haywire

TraGicCode: Done.

TraGicCode:
i'm going to try and add a few other docker sets to this before this gets merged.

TraGicCode:
@wyardley I added more docker nodesets as noted below: 
* centos-6
* centos-7
* debian-7
* debian-8

TraGicCode:
Closes #149

.sync.yml
```diff
@@ -1,6 +1,13 @@
 ---
 .travis.yml:
   secure: "RrAmWtM6eTjZZzD954AIgR179Pqp14lzHhd/C9cpKbTPynLncuim08CEvjmq+7pgAy9XDg1d02x3udfZt4btR1sBdyNRpNN3yUhWptmGU61HRJdiZq+nSLQkIYsqXanhk+O3NndFojO58WRD01dkWEc6HRHjlivuYNxDXmMkg9k="
+  docker_sets:
+  - set: docker/centos-6
+  - set: docker/centos-7
+  - set: docker/debian-7
+  - set: docker/debian-8
+  - set: docker/ubuntu-14.04
+  - set: docker/ubuntu-16.04
```
bastelfreak: can you split the new nodesets at least into an own commit? Or a separate PR.

TraGicCode: Sure thing!

PR #161, opened 2017-11-20
replace validate_string with assert_type

bastelfreak:

### DIFF
--- a/manifests/addon.pp
+++ b/manifests/addon.pp
@@ -32,16 +32,16 @@
 #   }
 # }
 #
+
 define splunk::addon (
-  $splunk_home    = '/opt/splunkforwarder',
-  $package_manage = true,
-  $package_name   = undef,
-  $inputs         = {},
+  $splunk_home                   = '/opt/splunkforwarder',
+  $package_manage                = true,
+  Optional[String] $package_name = undef,
+  $inputs                        = {},
 ) {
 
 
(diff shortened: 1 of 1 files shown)


manifests/addon.pp
```diff
@@ -32,16 +32,17 @@
...
+  $splunk_home                   = '/opt/splunkforwarder',
+  $package_manage                = true,
+  Optional[String] $package_name = undef,
+  $inputs                        = {},
 ) {
 
 
   if ( $package_manage ) {
-    validate_string($package_name)
+    assert_type(String, $package_name)
```
juniorsysadmin: Shouldn't the data typing already cover this?

bastelfreak: ups, copy and paste failure. let me fix that.

bastelfreak:
@juniorsysadmin fixed.

manifests/addon.pp
alexjfisher: It *was* also allowed to be `undef`.  But if `package_manage == true` then I think it probably does have to be a string.

juniorsysadmin: @alexjfisher is right. But now that I think of it, what you had was better before. That accounts for the case where someone specified `$package_name` = true but did not provide a proper `$package_name`. `undef` as a value for the package resource name is also invalid in Puppet 4/5 I believe.

PR #162, opened 2017-11-22
Update README

arjenz:
Update the text about the example to match the versions in the actual
example


bastelfreak:
Hi @arjenz, thanks for the PR. Can you take a look at the used email address in your commit? It isn't associated with your github account.

arjenz:
@bastelfreak It should be now.

bastelfreak:
Thanks!

Issue #125, opened 2017-08-31
Allow Splunk/Splunkforwarder to be installed with "package" resource


inspired:
This is an enhancement request.
For those using Satellite or a package manager with the splunkforwarder in their repos it would be great to be able to install the splunkforwarder with the package resource instead of bundling it with the module/in Puppet repo.
Any reason this is not supported?

bastelfreak:
hi @inspired, I'm not 100% sure what you mean. The package is currenty installed via package resource. Do you want to manage it outside of this module?

inspired:
@bastelfreak Yeah, I'd like to install it with a RPM or APT repo instead of bundling it with the module. As far as I can see that's not supported? Correct me if I'm wrong :-)

arjenz:
I'd like this feature too (was about to implement it myself, but haven't gotten around to it). Where possible, I try to add the Splunk packages to an internal repo, so yum and apt are able to just "simply" install them without puppet providing the binary package as a file.

wilrodriguez:
If we could even just add a flag for this class to simply not manage the package that would even solve the problem for me.

anthonysomerset:
if you set source to '' in the splunk class declaration, then it will assume to use package repos, we do this already internally and it works well for us

jstraw:
I would even be ok with the puppet module package installing the splunkforwarder package, I have requirements that all my packages be managed centrally in a repository owned by a compliance team.  So our firewalls would block access to download the rpms in other ways, but there's an existing repo with splunkforwarder available.

DLeich:
I'm going to add my +1 to this request.  Storing binaries and artifacts alongside source code in git is a really bad idea.  It's highly inefficient for git and wastes storage storage unnecessarily.  I am confused why we're not suggesting the use of the alternative source mechanism by default?
https://github.com/voxpupuli/puppet-splunk/blob/master/manifests/forwarder.pp#L49-L54

My understanding from reading the code is if you set $manage_package_source to false, and then provide a url to $package_source, then this should be doable.

PR #164, opened 2017-12-07
WIP: Add upgrade/version handling for splunkforwarder

arjenz:
This allows upgrading splunkforwarder by bumping version/build passed to this module. As a workaround more or less for https://github.com/voxpupuli/puppet-splunk/issues/125.

Changing the version/build parameters to `splunk::params` results in:
```
Notice: /Stage[main]/Splunk::Forwarder/Package[splunkforwarder]/ensure: ensure changed '6.5.1-f74036626f0c' to '6.6.2'
```

Only implemented for splunkforwarder and tested on CentOS right now. If there's interest I can expand it to support splunk and test on other distributions.

- Change "ensure" to an explicit version (otherwise only the file in /opt/staging will be added when $version is bumped)
- Extra handling of accepting the license

`ensure => absent` for the ftr file might even completely replace `Exec['license_splunkforwarder']` eventually (untested, however)


bastelfreak:
Hi @arjenz, thanks for the PR. Can you take a look at the failing travis jobs?

kenmaglio:
Just as a note, we've noticed the exec virtual resource won't be realized on an upgrade, as the file (/opt/splunkforwarder/etc/auth/server.pem) continues to exist between upgrades.  Since this modification isn't removing that file as well, then this would never allow the exec to fire.

PR #176, opened 2018-03-27
New Custom Type For Managing .meta files

michaelweiser:
This change enables the module to manage default.meta and local.meta
files in splunk system and app contexts. This allows to deploy access
control and export settings.

The patch introduces a new type splunk_metadata which behaves similar to
the other types and has the same interface. Internally however it
redirects to the metadata directory of the system or app contexts and
manages the files default.meta or local.meta respectively.

Example:
```puppet
splunk_metadata { 'app foo system export':
  section => '',
  setting => 'export',
  value   => 'system',
  context => 'app/foo/default',
  # -> $SPLUNK_HOME/etc/app/foo/metadata/default.meta
}

splunk_metadata { 'system access':
  section => '',
  setting => 'access',
  value   => 'read : [ me ], write : [ admin ]',
  # context default system local
  # -> $SPLUNK_HOME/etc/system/metadata/local.meta
}
```
Metadata files regularly have a section with an empty name ([])
containing the app access permissions. Since Puppet's IniFile class does
not support this directly, the provider subclasses
Puppet::Util::IniFile::Section and Puppet::Util::IniFile to add the
necessary functionality.

Additionally, the context parameter is made a namevar so that multiple
sections and settings with the same name can exist in different
contexts.


bastelfreak:
Hi @michaelweiser, thanks for the PR! Can you take a look at the failing spec tests?

michaelweiser:
I think those are related to the staged_package problem you addressed in https://github.com/voxpupuli/puppet-splunk/pull/179. They were happening with unmodified puppet-splunk when running the testsuite locally as well. I do not think they're related to my changes.

bastelfreak:
@michaelweiser yes correct, I just noticed this too. We're currently working on a fix in https://github.com/voxpupuli/puppet-splunk/pull/179. Please feel free to take a look.

Issue #169, opened 2018-01-29
Warning unknown variable 'staged_package' when `yum` is used as package provider


rmestrum:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet 4.10.8:
- Ruby: 2.1.9p490
- Distribution: RHEL 6.7
- Module version: 7.1.0

## How to reproduce (e.g Puppet code you use)

```
  class {'::splunk::forwarder':
    pkg_provider => 'yum',
  }
```

## What are you seeing
Puppet gives the following warning:
```
Warning: Unknown variable: 'staged_package'. at /etc/puppetlabs/code/environments/production/modules/splunk/manifests/forwarder.pp:93:28
```
## What behaviour did you expect instead
No warning.

bastelfreak:
Hi, thanks for rising this bug. Interesting that it is unknown. I am not quite sure how that is possible. Do you call any other class of the splunk module as well?

rmestrum:
I only use this class since we only install the splunk forwarder.

bastelfreak:
@TraGicCode can you take a look please? Any idea?

TraGicCode:
@bastelfreak I'm actually noticing this on my puppetserver as well ( which is ubuntu ).  We should be able to throw together some rspec-puppet tests to reproduce this and hopefully pinpoint and fix the bug.  Unfortunately i cannot look at this at the moment as things are very very busy for me with work since we have a huge migration project.  I can try and look at this when i get the chance.

TraGicCode:
It looks like the cause of this is seems to be happening in this giant mess.

https://github.com/voxpupuli/puppet-splunk/blob/26d93aeca889f6486e8a4f689e6cb36f76f7172f/manifests/forwarder.pp#L75-L95

1. The conditional there says if package provider is != yum then don't execute the code in the conditional. 
1. Below the conditional you see some package resource defaults being set.  It essentially says if the package provider is not chocolatey then assume the installer for splunk-forwarder is going to come from a downloaded file from the archive resource ( which didn't happen based on the conditional )

This whole mess really need to be cleaned up.  


Also inside the params.pp there is some code that really bothers me.  It's trying to pick the default package provider based on the node's o/s family fact.  This should work on deciding the "default" should be done automatically by the package provider implementations provided by puppet.  

[...]

PR #178, opened 2018-03-28
bump puppet to latest supported version 4.10.0

bastelfreak:


bastelfreak:
I'm wondering myself how that ever compiled. We're referencing the variable `$staged_package` in the init.pp but it isn't always available.

TraGicCode:
Maybe it’s time to fix #169 now :D

Issue #151, opened 2017-10-02
[Feature] - Ability to Install and Managed Splunk Server on Windows O/S


TraGicCode:
Currently this module does not work when installing splunk server on windows O/S.  Lets get this feature working!

Here is a bug that was reported because of this feature missing. #115

PR #179, opened 2018-03-28
fix unknown variable: staged_package

bastelfreak:

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -85,6 +85,8 @@ class splunk::forwarder (
       extract => false,
       before  => Package[$package_name],
     }
+  } else {
+    $staged_package = undef
   }
 
   Package  {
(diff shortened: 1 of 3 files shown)


bastelfreak:
This is a proposed fix for https://github.com/voxpupuli/puppet-splunk/issues/169. This reveals new errors that I don't quite understand yet. I'm not sure if we've multiple bugs in the code or if this patch is wrong. @TraGicCode and idea?

TraGicCode:
Yea i believe that was what was proposed.  It's more of a patch then a true fix of the actual problem.

TraGicCode:
This seems fine to me.  I think the these errors were hidden by this bug you just fixed.  Can i merge this?

TraGicCode:
Based on the changes #180 was trying to make i understand now why it's failing.  I reported this a while back #151.

PR #180, opened 2018-03-28
Windows fixes

michaelweiser:
Hi,

when rebased against https://github.com/voxpupuli/puppet-splunk/pull/179 these changes fix the remaining Windows problems:

- the config file resources actually do not use the correct path on Windows and this problem is called out by the test suite
- service and package names differ on Windows

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -272,77 +272,77 @@ class splunk (
     mode => '0600',
   }
 
-  file { '/opt/splunk/etc/system/local/alert_actions.conf':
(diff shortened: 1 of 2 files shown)


TraGicCode:
Hey @michaelweiser ,

I went ahead and simply excluded windows from the splunk-server tests.  Even fixing the unit tests would require more testing and this makes the most sense at the time.  I also noticed the packages you had for windows were different then what i had installed on my windows box for splunk server.  Anyways just giving you a heads up and will go ahead and close this.  If you Feel like a challenge free to implement #151 and add these tests :)

PR #183, opened 2018-06-12
Allow forwarder to fully support alternate install directories

Nekototori:
Ran into issues with getting the splunk forwarder installation working when installing it into a different directory.  splunkforwarder_web passes into splunk's ini_setting for determining the file_path, which pulls from splunk_config, but that isn't set within forwarder.  This pull just adds support for that.

The spec test for the forwarder class just has the compile with deps, so I didn't add anything to test values of the path for it.


bastelfreak:
Thanks for the PR @Nekototori !

Issue #159, opened 2017-10-29
Splunk as non-root user by default for UNIX systems.


jbethmont:
This is an enhancement request.
As part of Splunk Enterprise best practices Splunk should run as 'non-root user'. See official splunk documentation http://docs.splunk.com/Documentation/Splunk/7.0.0/Installation/RunSplunkasadifferentornon-rootuser

## Code involved: ~/puppet-splunk/manifests/params.pp

`  } else {
    $forwarder_dir = pick($forwarder_installdir, '/opt/splunkforwarder')
    $server_dir    = pick($server_installdir, '/opt/splunk')
    $splunk_user   = 'root'
  }`

## What are you seeing

splunk runs as "root" user by default. Which is against splunk best practices.

## What behaviour did you expect instead

The module should create a non-root user (i.e "splunk" user) and run splunk as it.

TraGicCode:
Hey @jbethmont ,
Thanks for the enhancement request. Here are a couple things we need to figure out first.

1.) What is the default chosen/selected user for splunk when fresh installed.  If this is wrong we should get this fixed up. If the default user on fresh install is not based on splunk best practices I would argue that this should be left alone.  I would think the scenario your describing is for advanced users and might be too restrictive for regular users to figure out.

jbethmont:
@TraGicCode The default user on a fresh install is the user that owns the splunk directory and is used to launch the splunk process. So technically it could be any user. But Splunk has the ability to run scripts as scripted inputs on the system. So for security best practices we don't want Splunk to run as root.

I would recommend then that the module creates a "splunk" user by default that will own the directory and run splunk processes. But definitely not "root" by default".

And for the module not to be "too restrictive", the user should be able to override the 'splunk_user' value.

PR #195, opened 2018-09-05
Make splunk_user configurable.

mdwheele:
Hello!

This PR addresses #159 without making any backwards-incompatible changes. I would like to be able to configure the Splunk user but have no opinion I wish to assert over what should be the default for this module. There are certainly best-practices from Splunk that advise running as `splunk`, but the default for a long time was to run as `Administrator` or `root`, as configured here by default. In 2018, this is no longer the case. However, I could totally see environments with extreme change control processes that MUST maintain current configs (and have likely implemented compensating controls to protect against the fact that Splunk has higher privilege than necessary).

All that said, I think enabling folks to make the call for their own environment is the best path forward and would appreciate your consideration!

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,6 +26,9 @@
 #   The port on which to send logs, and listen for logs. Used as a default for
 #   splunk and splunk::forwarder.
 #
+# [*splunk_user*]
+#   The user that splunk runs as.
+#
 # [*src_root*]
 #   The root URL at which to find the splunk packages. The sane-default logic
 #   assumes that the packages are located under this URL in the same way that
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -86,6 +89,10 @@
   $server               = 'splunk',
   $forwarder_installdir = undef,
   $server_installdir    = undef,
+  $splunk_user          = $::osfamily ? {
```
bastelfreak: can you please use ```facts$['os']['family']``` as fact here? The old toplevel variables are deprecated.

bastelfreak:
Can you take a look at the failing spec tests?

mdwheele:
It looks like it simply couldn't do the lookup in parameters list because `:facts` wasn't populated to match expectations of the new default assignment out of `$facts['os']['family']`. I think it's good to go now.

mdwheele:
No problem! I really appreciate all the work y'all have put into this module. It's very helpful.

PR #201, opened 2018-10-18
Allow empty local and outputs files to keep centralized Splunk happier.

gregswift:
Splunk environments where the majority of the configuration is handled by splunk do not necessarily like these defaults. So allow a way to not provide them. Either by setting undef or passing empty (nil) from hiera.

### DIFF
--- a/manifests/forwarder.pp
+++ b/manifests/forwarder.pp
@@ -118,10 +118,10 @@ class splunk::forwarder (
 
   # Declare inputs and outputs specific to the forwarder profile
   $tag_resources = { tag => 'splunk_forwarder' }
-  if $forwarder_input != undef {
+  if $forwarder_input {
     create_resources( 'splunkforwarder_input',$forwarder_input, $tag_resources)
   }
-  if $forwarder_output != undef {
+  if $forwarder_output {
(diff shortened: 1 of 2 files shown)


manifests/forwarder.pp
```diff
@@ -112,8 +118,12 @@
 
   # Declare inputs and outputs specific to the forwarder profile
   $tag_resources = { tag => 'splunk_forwarder' }
-  create_resources( 'splunkforwarder_input',$forwarder_input, $tag_resources)
-  create_resources( 'splunkforwarder_output',$forwarder_output, $tag_resources)
+  if $forwarder_input != undef {
```
bastelfreak:
```suggestion
  if $forwarder_input {
```

manifests/forwarder.pp
```diff
@@ -112,8 +118,12 @@
...
+    create_resources( 'splunkforwarder_input',$forwarder_input, $tag_resources)
+  }
+  if $forwarder_output != undef {
```
bastelfreak:
```suggestion
  if $forwarder_output  {
```

bastelfreak:
Thanks @gregswift !

PR #209, opened 2018-11-28
Add the ability to override package_source

beezly:
Allows you to specify manage_package_source to override the modules behaviour when you want to specify your own package_source. The existing functionality in the module is not enough, as sometimes it is necessary to specify an undefined value, (for example, if you are using yum to install from your own repos - the source parameter to the package type needs to be undef) which at the moment gets overridden by the value from $splunk::params::server_pkg_source


manifests/forwarder.pp
```diff
@@ -52,7 +56,8 @@
 #
 class splunk::forwarder (
   $server                 = $splunk::params::server,
-  $package_source         = $splunk::params::forwarder_pkg_src,
+  $manage_package_source  = true,
```
bastelfreak: Can you please add datatypes for those parameters?

manifests/init.pp
```diff
@@ -44,7 +48,8 @@
 # Requires: nothing
 #
 class splunk (
-  $package_source         = $splunk::params::server_pkg_src,
+  $manage_package_source  = true,
```
bastelfreak: Same here, can you please add dataypes?

bastelfreak:
Hi @beezly. thanks for the PR! Can you take a look at the inline comments I made and the failing spec tests?

beezly:
@bastelfreak I've added datatypes to those two classes and fixed the failing tests. It looks like one job is going to fail (https://travis-ci.org/voxpupuli/puppet-splunk/jobs/461227416) but I believe that is unrelated to the changes I've made (it's failing to restart the ssh service when setting up a debian 8 docker container to run the tests in). I'll give it a nudge and see if it passes a second time.

beezly:
Also - apologies for the close/reopen. It was the only way I could get Travis to re-run the tests after the failed run earlier.

manifests/forwarder.pp
```diff
@@ -51,43 +55,50 @@
...
-  $forwarder_output       = $splunk::params::forwarder_output,
-  $forwarder_input        = $splunk::params::forwarder_input,
-  $create_password        = $splunk::params::create_password,
-  $addons                 = {},
+  String $server                        = $splunk::params::server,
+  Boolean $manage_package_source        = true,
+  Optional[String] $package_source      = undef,
+  String $package_name                  = $splunk::params::forwarder_pkg_name,
+  String $package_ensure                = $splunk::params::forwarder_pkg_ensure,
+  Variant[String,Integer] $logging_port = $splunk::params::logging_port,
```
alexjfisher: I wouldn't be against being more aggressive with these type changes even if it'd mean a breaking change.
eg using `Stdlib::Port` here.

alexjfisher: @bastelfreak Thoughts?

alexjfisher:
> Also - apologies for the close/reopen. It was the only way I could get Travis to re-run the tests after the failed run earlier.

That's fine.  I think it's even documented as a way to rerun tests in one of @bastelfreak's presentations.

manifests/forwarder.pp
TraGicCode: @alexjfisher I vote for doing it.

bastelfreak: :+1:, please add stricter types.

bastelfreak:
@alexjfisher @beezly correct, closing/reopening is totally fine.

manifests/forwarder.pp
beezly: I've added stricter types where I thought it was appropriate and updated that weird params class to have types set there too.

manifests/forwarder.pp
```diff
@@ -51,43 +55,50 @@
...
+  Boolean $create_password              = $splunk::params::create_password,
+  Hash $addons                          = {},
 ) inherits splunk::params {
 
   $virtual_service = $splunk::params::forwarder_service
   $staging_subdir  = $splunk::params::staging_subdir
 
   $path_delimiter  = $splunk::params::path_delimiter
+
+  $_package_source = $manage_package_source ? {
```
alexjfisher: Maybe make this a simple `if` and check `package_source` isn't `undef` when `manage_package_source` is `false`.

Something along the lines of
```puppet
if $manage_package_source {
  $_package_source = $splunk::params::forwarder_pkg_src
} else {
  unless $package_source { fail('`package_source` parameter must be set when `manage_package_source` is false') }
  $_package_source = $package_source
}
```
?

alexjfisher: ah, ok.  Didn't at first spot the `if $pkg_provider != undef and $pkg_provider != 'yum' and $pkg_provider != 'apt' and $pkg_provider != 'chocolatey'`

PR #214, opened 2019-02-05
WIP: Add 7.2.X support

### LINKED ISSUE #213: Module does not work with Splunk 7.1.x or 7.2.x
In Splunk 7.1 and above, Splunk has removed the splunk service. 

- Puppet: 4.10.12
- Ruby: ruby 2.0.0p648
- Distribution: RHEL 7
- Module version: 4fe1532433d406df26bd66d819da10c5c6373aa2

With the Splunk 7.2.3 RPM in a yum repository accessible to your system:

```
  class { '::splunk':
    pkg_provider => 'yum',
    splunk_user  => 'splunk',
  }
```

## What are you seeing

Splunk fails due to the splunk service no longer existing in Splunk 7.1 and 7.2. It seems that on Linux the new service setup might look a lot more like the windows setup, with there only being 2 services, `Splunkd`, and `splunkweb`.

A successful install of Splunk

## Output log
```
Error: Systemd start for splunk failed!
journalctl log for splunk:
-- No entries --

Error: /Stage[main]/Splunk::Virtual/Service[splunk]/ensure: change from stopped to running failed: Systemd start for splunk failed!
journalctl log for splunk:
-- No entries --
```

It seems that splunk has also renamed the `splunkd` service to `Splunkd`.

nick-markowski:
* Systemd, boot-start, and legacy mode support
  * Run the proper service script or binary commands when in legacy vs non-boot-start vs normal mode
  * Default to using system service file unless boot-start is disabled
  * Set appServerPorts to 0 when in legacy mode
* Added spec for Linux and SunOS kernels, updated acceptance
* Bumped stdlib dependency, strong typing requires >= 4.25.0
* Added OEL to supported os

Fixes #213


Gemfile
```diff
@@ -43,7 +43,7 @@ group :system_tests do
   if beaker_version = ENV['BEAKER_VERSION']
     gem 'beaker', *location_for(beaker_version)
   else
-    gem 'beaker', '>= 3.9.0', :require => false
+    gem 'beaker', '>= 4.2.0', :require => false
```
bastelfreak: can you please rebase against our master? Since https://github.com/voxpupuli/puppet-splunk/pull/211 got merged, those changes to the Gemfile shouldn't be needed.

nick-markowski:
@bastelfreak Thanks for the quick feedback :) It seems a lot of the nodesets are missing the master role - is my test wrong, or are the nodesets from the recent modulesync out of whack?

spec/acceptance/splunk_spec.rb
```diff
@@ -17,7 +20,15 @@ class { '::splunk': }
       it { is_expected.to be_installed }
     end
 
-    describe service('splunk') do
+    service_provider = fact_on(splunk_host, 'service_provider')
+    case service_provider
```
nick-markowski: Needs to be adjusted per splunk version

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-splunk",
-  "version": "7.3.1-rc0",
+  "version": "7.3.1",
```
bastelfreak: please don't adjust the version. We will do this during the next release.

nick-markowski: Woops! Forgot to replace the rc.

nick-markowski:
@bastelfreak I believe setting boot_start true (the current default) is a breaking change - the previous version of the module enabled the service but used the binaries instead of the service file to manage the splunk service.  Upgrading to this version may cause a new instance of splunk to start up (because the service provider will return that the service is inactive/dead), which is bad.  I can default boot_start to false for a seamless upgrade, or we can bump the default version back to 7.2.3, enable boot_start, and release a major update to account for the breaking changes.  Thoughts?

bastelfreak:
Hi,
should we implement this as  a non-breaking change, do a release and afterwards update the defaults and do a new major release?

bastelfreak:
ha, crap. We already merged https://github.com/voxpupuli/puppet-splunk/pull/211, which is a breaking change. IMO it's fine if this PR os another breaking change, it just needs to be documented properly in the readme.

nick-markowski:
Heh, ok. I'll leave the defaults the way they currently are so boot_start is default, but I won't upgrade the splunk version.  I have not tested a splunk upgrade with this module and honestly don't care to for the scope of this change.  I can file a new feature request and address it later on.

nick-markowski:
I need more time to test these changes, I may reverse course on some of my decisions. I'm not entirely sure what the effect of toggling options such as boot_start or legacy mode would do, because there's not a clear path to re-enable the service without potentially causing damage.  I may have been a bit too ambitious supporting those modes. Marking as WIP until I can sort it out.

bastelfreak:
okay, let us know when you are done.

nick-markowski:
@bastelfreak after going down the rabbit hole of gottchas with toggling legacy mode and boot start mode, I've decided it's probably best to drop legacy mode support.  I'm going to re-write my pr and get it tested/reviewed hopefully by the end of the week. Thanks for your patience :)

nick-markowski:
Decided the module needs a larger refactor to account for 7.2 updates and modulesync.  Closing out until that refactor is done.

PR #215, opened 2019-03-05
Add 7.2 support and refactor module

### LINKED ISSUE #210: Unable to set user for forwarder service
It would be helpful if we could set the user used for running the splunk forwarder service

### LINKED ISSUE #212: Unable to create symlink

- Puppet: 1.10.14-1stretch (for example)
- Distribution: Debian 9
- Module version: 7.3.0

## Description
Splunk versions above 7.2.1 don't create file /etc/init.d/splunk during the command `splunk enable boot-start -user ${splunk_user}`. It creates symlink on systemd unit (service):

```bash
ls -l /etc/systemd/system/multi-user.target.wants/SplunkForwarder.service 
lrwxrwxrwx 1 root root 43 Jan 16 13:30 /etc/systemd/system/multi-user.target.wants/SplunkForwarder.service -> /etc/systemd/system/SplunkForwarder.service
```

But in `manifests/platform/posix.pp` execs expect that file `/etc/init.d/splunk` is created. So on every run of puppet agent, agent tries to create this symlink and gets an error:

```bash
Notice: /Stage[main]/Splunk::Platform::Posix/Exec[enable_splunkforwarder]/returns: Unable to create symlink='/etc/systemd/system/multi-user.target.wants/SplunkForwarder.service'
Notice: /Stage[main]/Splunk::Platform::Posix/Exec[enable_splunkforwarder]/returns: : File exists
Error: '/opt/splunkforwarder/bin/splunk enable boot-start -user root' returned 8 instead of one of [0]
Error: /Stage[main]/Splunk::Platform::Posix/Exec[enable_splunkforwarder]/returns: change from notrun to 0 failed: '/opt/splunkforwarder/bin/splunk enable boot-start -user root' returned 8 instead of one of [0]
```

I [commited](https://github.com/nikitashalnov/puppet-splunk/commit/eb6b794f29c7843d23b062541c32cecd2f87ef0f) some changes to my fork, but not sure that this is the proper way to fix this problem.

### LINKED ISSUE #213: Module does not work with Splunk 7.1.x or 7.2.x

nick-markowski:
* Supports systemd boot-start
* Changed module design from Class[splunk], Class[splunk::forwarder] to
  Class[splunk::enterprise] and Class[splunk::forwarder]
* Removed legacy-style service management
* Allow users to disable default config (splunk/forwarder_* types)
* Removed ability to include enterprise and forwarder on the same machine
  * Forwarder is simply a subset of enterprise capability and there is no need
    to have both
  * Forwarder and enterprise potentially share common service names and other
    assets, which cause issues when including both
* Bumped dependency on stdlib to 4.25.0 to ensure strong typing support
* Allow users to specify splunk service user
* Added password management to enterprise

Fixes #213
Fixes #212
Fixes #210


nick-markowski:
Still need to:
* Update in-module and README parameter documentation once design is approved
* Update acceptance tests
* Optionally upgrade test, and default the version to the latest release

manifests/addon.pp
```diff
@@ -34,28 +34,36 @@
...
 
 define splunk::addon (
-  $splunk_home                   = '/opt/splunkforwarder',
+  $splunk_home                   = undef,
   $package_manage                = true,
   Optional[String] $package_name = undef,
   $inputs                        = {},
 ) {
 
+  include '::splunk::params'
```
bastelfreak: please remove the leading `::`. That is legacy syntax that was required for puppet 3.

manifests/addon.pp
```diff
@@ -34,28 +34,36 @@
...
+
+  if !defined($splunk_home) {
+    $_splunk_home = $::splunk::params::forwarder_homedir
```
bastelfreak: same here, please remove the leading `::`.

manifests/enterprise.pp
```diff
@@ -0,0 +1,82 @@
+# Class splunk::enterprise
+#
+class splunk::enterprise (
+  String $version                          = $::splunk::params::version,
```
bastelfreak: for all the strings, can you update it to `String[1]` to prohibit empty strings?

manifests/addon.pp
```diff
@@ -34,28 +34,36 @@
...
-  $splunk_home                   = '/opt/splunkforwarder',
+  $splunk_home                   = undef,
   $package_manage                = true,
   Optional[String] $package_name = undef,
   $inputs                        = {},
 ) {
 
+  include '::splunk::params'
+
+  if !defined($splunk_home) {
```
alexjfisher: or just `unless $splunk_home {`

alexjfisher: Or `if $splunk_home` and reverse the blocks.

alexjfisher:
@nick-markowski Huge thanks for taking this on!  It's something I'm going to be needing myself shortly, so I should be able to test your PR sometime over the next week or so.

manifests/enterprise.pp
```diff
@@ -0,0 +1,82 @@
...
+  Optional[String] $package_source         = undef,
+  Optional[Array[String]] $install_options = $::splunk::params::enterprise_install_options,
+  String $splunk_user                      = $::splunk::params::splunk_user,
+  String $enterprise_homedir               = $::splunk::params::enterprise_homedir,
+  String $enterprise_confdir               = $::splunk::params::enterprise_confdir,
+  String $service_name                     = $::splunk::params::enterprise_service,
+  String $service_file                     = $::splunk::params::enterprise_service_file,
+  Boolean $boot_start                      = $::splunk::params::boot_start,
+  Boolean $use_default_config              = true,
+  String $input_default_host               = $facts['clientcert'],
```
alexjfisher: `clientcert`?  That doesn't actually have to be a hostname.  Probably better to stick with `$facts['fqdn']` here.

manifests/enterprise.pp
```diff
@@ -0,0 +1,82 @@
...
+  String $version                          = $::splunk::params::version,
+  String $package_name                     = $::splunk::params::enterprise_package_name,
+  String $package_ensure                   = $::splunk::params::enterprise_package_ensure,
+  String $staging_subdir                   = $::splunk::params::staging_subdir,
+  String $path_delimiter                   = $::splunk::params::path_delimiter,
+  String $enterprise_package_src           = $::splunk::params::enterprise_package_src,
+  Optional[String] $package_provider       = $::splunk::params::package_provider,
+  Boolean $manage_package_source           = true,
+  Optional[String] $package_source         = undef,
+  Optional[Array[String]] $install_options = $::splunk::params::enterprise_install_options,
```
alexjfisher: Instead of this being `Optional`, can we default to an empty array instead?

manifests/enterprise.pp
```diff
@@ -0,0 +1,82 @@
...
+  String $splunk_user                      = $::splunk::params::splunk_user,
+  String $enterprise_homedir               = $::splunk::params::enterprise_homedir,
```
alexjfisher: `Stdlib::Absolutepath` instead of `String`?

manifests/addon.pp
nick-markowski: I have...a lot of these to remove lol

manifests/enterprise.pp
```diff
@@ -0,0 +1,82 @@
...
+  String $enterprise_confdir               = $::splunk::params::enterprise_confdir,
+  String $service_name                     = $::splunk::params::enterprise_service,
+  String $service_file                     = $::splunk::params::enterprise_service_file,
+  Boolean $boot_start                      = $::splunk::params::boot_start,
+  Boolean $use_default_config              = true,
+  String $input_default_host               = $facts['clientcert'],
+  String $input_connection_host            = 'dns',
+  String $splunkd_listen                   = '127.0.0.1',
```
alexjfisher: `Stdlib::IP::Address`?

nick-markowski: I need to comb through all of the strong typing, I winged it

manifests/enterprise.pp
nick-markowski: sure can!

nick-markowski:
@alexjfisher @bastelfreak Splunk has managed to screw up service management once again, and have made an assumption that cgroups are present in the unit file, if enabled (boot-start).  Cgroups are obviously not present in containers, so the test suite fails:

chown: cannot access ‘/sys/fs/cgroup/cpu/system.slice/Splunkd.service’: No such file or directory

One way we can probably get around it is to hack in the 2 cgroup files the until file expects to manage.  It's ugly, but i'm not sure what else to do until Splunk fixes their code.

manifests/enterprise/install.pp
```diff
@@ -0,0 +1,46 @@
...
+    include splunk::enterprise::install::nix
+  }
+
+  $_package_source = $splunk::enterprise::manage_package_source ? {
+    true  => $splunk::enterprise::enterprise_package_src,
+    false => $splunk::enterprise::package_source
+  }
+
+  if $splunk::enterprise::package_provider != undef and $splunk::enterprise::package_provider != 'yum' and $splunk::enterprise::package_provider != 'apt' and $splunk::enterprise::package_provider != 'chocolatey' {
+    include ::archive::staging
```
bastelfreak: please use:
```suggestion
    include archive::staging
```

ody:
Hey @nick-markowski,

How can I help get this over the line and merged? Working on a project and officially blocked by the lack of 7.2 support.

nick-markowski:
Hey @ody !  I have a list of todo items I should be able to begin addressing this week:

* Update in-module and README parameter documentation once design is approved
  * I need to slog through copying the old parameter docs to the new classes and ensure new parameters are documented
* Update acceptance tests to test systemd/backwards compatibility changes thoroughly
  * This is a good place for input! I can only manually test so much, and the test suite needs upgrading
  * Test add-on and password if possible, I don't have a good use case for either
  * Optionally upgrade test, and default the version to the latest release (already done, but we need to update the test suite)

Before the tests can be completed, I need some resolution about how to handle issues I posted in the chain, above.

alexjfisher:
> @alexjfisher @bastelfreak Splunk has made an assumption that cgroups are present in the systmed unit file, if enabled (boot-start). Cgroups are obviously not present in containers, so the test suite fails:
> 
> chown: cannot access ‘/sys/fs/cgroup/cpu/system.slice/Splunkd.service’: No such file or directory
> 
> One way we can probably get around it is to hack in the 2 cgroup files the until file expects to manage. It's ugly, but i'm not sure what else to do until Splunk fixes their code.
> 
> EDIT: If you're running splunk in a container IRL you would probably not use systemd to manage the service, you would use the binaries directly to kick off the container (or download the splunk container, per their docs). We may need to update the travis tests to utilize virtual hosts instead of containers to fully test the suite.

Umm... I really don't know.  I wonder if @ekohl has any ideas?

ekohl:
> One way we can probably get around it is to hack in the 2 cgroup files the until file expects to manage. It's ugly, but i'm not sure what else to do until Splunk fixes their code.

This sounds fair. The spec_acceptance_helper sounds like the correct place if they just need to exist and setting it up once is enough.

> We may need to update the travis tests to utilize virtual hosts instead of containers to fully test the suite.

AFAIK we have no place to run VMs

nick-markowski:
@ekohl I'll try to get the tests running in the containers and report any other blockers, thanks!

nick-markowski:
@alexjfisher @bastelfreak looks like some dependencies within the debian nodeset are defunct, not sure how you want to handle it

ekohl:
Debian 8 is EOL. Technically it's LTS, but in Debian that has a different meaning. It's fine to drop Debian 8 as a supported OS and only support Debian 9 IMHO.

bastelfreak:
@nick-markowski are you happy with this as it is now? I don't want to review such a huge PR multiple times :D. Please remove the `WIP` from the title if this is ready.

nick-markowski:
@bastelfreak understandable! I left it WIP because I need to update the README and in-module param documentation.  I'll get that done this week.

manifests/addon.pp
```diff
@@ -34,28 +34,36 @@
 #
 
 define splunk::addon (
-  $splunk_home                   = '/opt/splunkforwarder',
-  $package_manage                = true,
-  Optional[String] $package_name = undef,
-  $inputs                        = {},
+  Optional[Stdlib::Absolutepath] $splunk_home = undef,
+  Boolead $package_manage                     = true,
```
juniorsysadmin: Boolean

nick-markowski:
@alexjfisher @bastelfreak @ekohl Looks like the review has already started and I need to add some testing - could I get some more 👀 on this? :)

alexjfisher:
@nick-markowski I'm going to be trying this change out for real in the next few days. :)

.sync.yml
```diff
@@ -4,7 +4,7 @@
   docker_sets:
   - set: centos6-64
   - set: centos7-64
-  - set: debian8-64
+  - set: debian9-64
```
bastelfreak: to make the whole diff smaller, do you think it's possible to add debian9 in a seperate PR and drop debian 8 in a seperate PR?

manifests/enterprise/config.pp
```diff
@@ -0,0 +1,149 @@
...
+    file { $splunk::enterprise::secret_file:
+      ensure  => file,
+      owner   => $splunk::enterprise::splunk_user,
+      group   => $splunk::enterprise::splunk_user,
+      content => $splunk::enterprise::secret,
+    }
+  }
+
+  # Remove init.d file if the service provider is systemd
+  if $facts['service_provider'] == 'systemd' and versioncmp($splunk::enterprise::version, '7.2.2') >= 0 {
```
bastelfreak: do we need the `versioncmp()` here? Isn't it enough to check for the `service_provider`?

manifests/enterprise/config.pp
```diff
@@ -0,0 +1,149 @@
...
+    file { '/etc/init.d/splunk':
+      ensure => 'absent',
+    }
+  }
+
+  if $facts['virtual'] == 'docker' {
+    ini_setting { 'OPTIMISTIC_ABOUT_FILE_LOCKING':
```
bastelfreak: lol, nice name for an option :D

manifests/enterprise/config.pp
```diff
@@ -0,0 +1,149 @@
...
+    }
+    splunk_web { 'splunk_server_web_port':
+      section => 'settings',
+      setting => 'httpport',
+      value   => $splunk::enterprise::web_httpport,
+      tag     => 'splunk_server',
+    }
+  }
+
+  File <| tag == 'splunk_enterprise' |>
```
bastelfreak: I'm wondering if we can simplify the whole block of class dependencies :thinking:

manifests/enterprise/install.pp
```diff
@@ -0,0 +1,45 @@
...
+  if $facts['kernel'] == 'Linux' or $facts['kernel'] == 'SunOS' {
+    include splunk::enterprise::install::nix
+  }
+
+  $_package_source = $splunk::enterprise::manage_package_source ? {
+    true  => $splunk::enterprise::enterprise_package_src,
+    false => $splunk::enterprise::package_source
+  }
+
+  if $splunk::enterprise::package_provider != undef and $splunk::enterprise::package_provider != 'yum' and $splunk::enterprise::package_provider != 'apt' and $splunk::enterprise::package_provider != 'chocolatey' {
```
bastelfreak: `if $splunk::enterprise::package_provider != undef` is equal to `if $splunk::enterprise::package_provider`

---

`if $splunk::enterprise::package_provider != 'yum' and $splunk::enterprise::package_provider != 'apt' and $splunk::enterprise::package_provider != 'chocolatey'` should be equal to `if not $splunk::enterprise::package_provider in ['apt', 'chocolatey', 'yum']` (but I didn't test it).

manifests/enterprise/install.pp
```diff
@@ -0,0 +1,45 @@
...
+    include archive::staging
```
bastelfreak: is this include actually required? :thinking:

manifests/enterprise/install/nix.pp
```diff
@@ -0,0 +1,41 @@
...
+    # Collect any Splunk packages and give them an admin and response file.
+    Package[$splunk::enterprise::enterprise_package_name] {
+      adminfile    => $_adminfile,
+      responsefile => $_responsefile,
+    }
+  }
+
+  # Required for splunk 7.2.4.2
+  if versioncmp($splunk::enterprise::version, '7.2.4.2') >= 0 {
+    if !defined(Package['net-tools']) {
```
bastelfreak: this could be simplified by using https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages

manifests/enterprise/service.pp
```diff
@@ -0,0 +1,19 @@
+# Class splunk::enterprise::service
+#
+class splunk::enterprise::service {
+
+  # This is a module that supports multiple platforms. For some platforms
+  # there is non-generic configuration that needs to be declared in addition
+  # to the agnostic resources declared here.
+  if $facts['kernel'] == 'Linux' or $facts['kernel'] == 'SunOS' {
```
bastelfreak: Untetested but should work and is shorter:

```suggestion
  if $facts['kernel'] in ['Linux', 'SunOS'] {
```

manifests/forwarder/install.pp
```diff
@@ -0,0 +1,68 @@
+# Class splunk::forwarder::install
+#
+class splunk::forwarder::install {
+
+  $_package_source = $splunk::forwarder::manage_package_source ? {
+    true  => $splunk::forwarder::forwarder_package_src,
+    false => $splunk::forwarder::package_source
+  }
+
+  if $splunk::forwarder::package_provider != undef and $splunk::forwarder::package_provider != 'yum' and $splunk::forwarder::package_provider != 'apt' and $splunk::forwarder::package_provider != 'chocolatey' {
```
bastelfreak: same here as mentioned above

manifests/enterprise/config.pp
nick-markowski: Splunk did not support systemd until 7.2.2, so enabling bootstart will drop a sysvinit file

.sync.yml
nick-markowski: I could re-introduce those changes as a separate PR (unless someone else beats me to it) and rebase this one to clean things up, for sure

manifests/enterprise/config.pp
nick-markowski: Lol, most of this is reused code I didn't find reason to modify :)

manifests/enterprise/install.pp
nick-markowski: I'll try cleaning this up, I have a local centos7 build I'm using yum providers with

manifests/enterprise/install.pp
nick-markowski: Good question, most of the crazy install logic I just forklifted and plopped into place...

manifests/enterprise/config.pp
nick-markowski: I hope so, it was painful to write these out! I guess, technically, we could order the types to consolidate them into one block; order is arbitrary anyways:

File <| tag == 'splunk_enterprise' |>
-> Splunk_authentication   <| tag == 'splunk_enterprise' |>
-> Splunk_authorize        <| tag == 'splunk_enterprise' |>
-> Splunk_deploymentclient <| tag == 'splunk_enterprise' |>
... etc
~> Class['splunk::enterprise::service']

manifests/forwarder.pp
```diff
@@ -1,153 +1,226 @@
...
-    section => 'settings',
-    setting => 'mgmtHostPort',
-    value   => "${splunkd_listen}:${splunkd_port}",
-    tag     => 'splunk_forwarder',
-  }
-
-  # If the purge parameters have been set, remove all unmanaged entries from
-  # the respective config files.
+# Class splunk::forwarder
+#
```
alexjfisher: nit.  Trailing whitespace on every other line.

alexjfisher: Not sure why puppet-lint doesn't complain.

manifests/forwarder.pp
```diff
@@ -1,153 +1,226 @@
...
+  String[1] $secret                          = $splunk::params::secret,
+  Hash $addons                               = {},
+) inherits splunk {
+
+  if (defined(Class['splunk::enterprise'])) {
+    fail('Splunk Universal Forwarder provides a subset of Splunk Enterprise capabilities, and has potentially conflicting resources when included with Splunk Enterprise on the same node.  Do not include splunk::forwarder on the same node as splunk::enterprise.  Configure Splunk Enterprise to meet your forwarding needs.'
+    )
+  }
+
+  include 'splunk::forwarder::install'
```
alexjfisher: Should these use `contain`?

eg what if I'm doing something like...

```puppet

pulp_rpmbind {'internal-splunk-7': }
class { 'splunk::forwarder':
  server                => $server,
  manage_package_source => false,
  package_provider      => 'yum',
  package_ensure        => 'latest',
  require               => Pulp_rpmbind['internal-splunk-7'],
}
```

README.md
```diff
@@ -316,177 +331,375 @@ The name of the package(s) Puppet will use to install Splunk.
...
+splunk::params.
+
+#### `secret`
+
+The secret used to salt the splunk password.  Defaults to the value set in
+splunk::params.
+
+#### `addons`
+
+Manage a splunk addons, see `splunk::addons`.  Defaults to an empty Hash.
```
alexjfisher: nit.  Should read `Manage splunk addons` (without the `a`).

manifests/forwarder/config.pp
```diff
@@ -0,0 +1,94 @@
...
+      tag     => 'splunk_forwarder',
+    }
+  }
+
+  # Declare addons
+  create_resources('splunk::addon', $splunk::forwarder::addons)
+
+  # Declare inputs and outputs specific to the forwarder profile
+  $_tag_resources = { tag => 'splunk_forwarder' }
+  if $splunk::forwarder::forwarder_input {
```
alexjfisher: This will *always* be truthy as `$splunk::forwarder::forwarder_input` is of type `Hash`.

manifests/forwarder/config.pp
```diff
@@ -0,0 +1,94 @@
...
+    create_resources( 'splunkforwarder_input', $splunk::forwarder::forwarder_input, $_tag_resources)
```
alexjfisher: You could/should use iteration instead of `create_resources`.

.sync.yml
ody: Is a move to Debian 9 a requirement for Splunk 7.2? That would be the only reason to rebase. If it isn't a requirement then it should be able to be handled entirely separately.

nick-markowski: I only updated because the debian 8 docker images were defunct and not letting the test suite pass, I'm fine with ditching the change and ignoring the deb8 tests

manifests/enterprise/install.pp
alexjfisher: This module is listed as supporting Windows though.

manifests/forwarder.pp
nick-markowski: Maybe comments are linted differently?

manifests/enterprise/config.pp
bastelfreak: @alexjfisher @binford2k can you have a look please?

ody:
Basic functionality of standing up a new instance of 7.2 and installing forwarder validated

Thanks @nick-markowski!

manifests/enterprise/service/nix.pp
```diff
@@ -1,10 +1,9 @@
 # Class splunk::enterprise::service::nix
 #
 class splunk::enterprise::service::nix inherits splunk::enterprise::service {
-#user, service, bootstart, homedir
 
   if $splunk::enterprise::boot_start {
-    # Eensure splunk services *not* managed by the system service file are
+    # Ensuure splunk services *not* managed by the system service file are
```
alexjfisher: So close! ;)

alexjfisher:
> Basic functionality of standing up a new instance of 7.2 and installing forwarder validated
> 
> Thanks @nick-markowski!

@ody Thanks for testing!  It's been looking good for me too.

manifests/forwarder/config.pp
```diff
@@ -0,0 +1,110 @@
...
+
+  # Declare addons
+  create_resources('splunk::addon', $splunk::forwarder::addons)
+
+  File <| tag == 'splunk_forwarder' |>
+  -> Splunkforwarder_deploymentclient <| tag == 'splunk_forwarder' |>
+  ~> Class['splunk::forwarder::service']
+
+  File <| tag == 'splunk_forwarder' |>
+  -> Splunkforwarder_input            <| tag == 'splunk_forwarder' |>
```
alexjfisher: What's the deal with only looking for resources with the specific tag?
Previously, you could declare these resources without a tag and they would notify the service.
See https://github.com/voxpupuli/puppet-splunk/pull/50

manifests/enterprise/config.pp
alexjfisher: > Removed ability to include enterprise and forwarder on the same machine

Was this the only reason for tagging the resources?  Maybe much of this could be simplified??

PR #153, opened 2017-10-09
Fix Idempotence issue for RHEL 7 forwarder install

TraGicCode:
I could really use a second pair of 👀 on this.  I was sifting through the issues on github for this module and noticed someone having a problem a RHEL 7.  While i was not able to reproduce their issue specifically ( i assume it's fixed ) i did notice the idempotence test for the forwarder is failing when run under beaker.  I know pretty much nothing about se_linux on RHEL 7 but what i do know is the following attribute on the file resource solves the forwarder idempotence test.  Could anyone verify this fix is indeed valid?

FYI.  Here is the failing beaker test without this change.

First Run
```
  Info: Loading facts
  Info: Loading facts
  Warning: /etc/puppetlabs/puppet/hiera.yaml: Use of 'hiera.yaml' version 3 is deprecated. It should be converted to version 5
     (in /etc/puppetlabs/puppet/hiera.yaml)
  Notice: Compiled catalog for rhel-74-x64.attlocal.net in environment production in 0.46 seconds
  Info: Applying configuration version '1507523793'
  Notice: /Stage[main]/Archive::Staging/File[/opt/staging]/ensure: created
  Notice: /Stage[main]/Splunk::Forwarder/Archive[/opt/staging/splunk/splunkforwarder-7.0.0-c8a78efdd40f-linux-2.6-x86_64.rpm]/ensure: download archive from https://download.splunk.com/products/universalforwarder/releases/7.0.0/linux/splunkforwarder-7.0.0-c8a78efdd40f-linux-2.6-x86_64.rpm to /opt/staging/splunk/splunkforwarder-7.0.0-c8a78efdd40f-linux-2.6-x86_64.rpm  with cleanup
  Notice: /Stage[main]/Splunk::Forwarder/Package[splunkforwarder]/ensure: created
  Notice: /Stage[main]/Splunk::Forwarder/Splunkforwarder_input[default_host]/ensure: created
  Info: /Stage[main]/Splunk::Forwarder/Splunkforwarder_input[default_host]: Scheduling refresh of Service[splunk]
  Notice: /Stage[main]/Splunk::Forwarder/Splunkforwarder_output[tcpout_defaultgroup]/ensure: created
  Info: /Stage[main]/Splunk::Forwarder/Splunkforwarder_output[tcpout_defaultgroup]: Scheduling refresh of Service[splunk]
  Notice: /Stage[main]/Splunk::Forwarder/Splunkforwarder_output[defaultgroup_server]/ensure: created
[...]

PR #221, opened 2019-04-06
Compare decrypted values to see if they are insync

alexjfisher:
When splunk is started, it automatically encrypts certain values when it
finds them in config files.  To stop puppet reverting these changes,
I've overriden `insync?` so that it performs the decryption before
comparing.

Currently only implemented for splunk >= 7.2

Based on description of algorithm in this python based project.
https://github.com/HurricaneLabs/splunksecrets

### DIFF
--- a/lib/puppet_x/voxpupuli/splunk/util.rb
+++ b/lib/puppet_x/voxpupuli/splunk/util.rb
@@ -1,0 +1,32 @@
+require 'openssl'
+require 'base64'
+
+module PuppetX
+  module Voxpupuli
+    module Splunk
+      class Util
+        def self.decrypt(secrets_file, value)
+          return value unless value.start_with?('$7$')
+
+          Puppet.debug "Decrypting splunk >= 7.2 data using secret from #{secrets_file}"
+          value.slice!(0, 3)
+          data = Base64.strict_decode64(value)
+          splunk_secret = IO.binread(secrets_file).chomp
+
+          iv         = data.bytes[0, 16].pack('c*')
+          tag        = data.bytes[-16..-1].pack('c*')
+          ciphertext = data.bytes[16..-17].pack('c*')
+
+          decipher = OpenSSL::Cipher::AES.new(256, :GCM).decrypt
+          decipher.key = OpenSSL::PKCS5.pbkdf2_hmac(splunk_secret, 'disk-encryption', 1, 32, OpenSSL::Digest::SHA256.new)
+          decipher.iv_len = 16
+          decipher.iv = iv
+          decipher.auth_tag = tag
+          decipher.auth_data = ''
+
+          decipher.update(ciphertext) + decipher.final
+        end
+      end
+    end
+  end
+end
--- a/spec/acceptance/splunk_enterprise_spec.rb
+++ b/spec/acceptance/splunk_enterprise_spec.rb
@@ -1,6 +1,13 @@
 require 'spec_helper_acceptance'
 
 describe 'splunk enterprise class' do
+  init = shell('/bin/readlink /sbin/init', acceptable_exit_codes: [0, 1]).stdout
+  service_name = if init.include? 'systemd'
+                   'Splunkd'
+                 else
+                   'splunk'
+                 end
+
   context 'default parameters' do
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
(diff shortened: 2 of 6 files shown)


lib/puppet_x/voxpupuli/splunk/util.rb
```diff
@@ -0,0 +1,32 @@
...
+          ciphertext = data.bytes[16..-17].pack('c*')
+
+          decipher = OpenSSL::Cipher::AES.new(256, :GCM).decrypt
+          decipher.key = OpenSSL::KDF.pbkdf2_hmac(splunk_secret, salt: 'disk-encryption', iterations: 1, length: 32, hash: 'sha256')
+          decipher.iv_len = 16
+          decipher.iv = iv
+          decipher.auth_tag = tag
+          decipher.auth_data = ''
+
+          decipher.update(ciphertext) + decipher.final
```
alexjfisher: Dunno if I should catch exceptions here, log a warning, and return something so that puppet can continue??

jorhett:
Quick question: would it be possible for Puppet to encrypt the values in a way that Splunk finds acceptable, so as to be more declarative / less subject to changes in Splunk?

lib/puppet_x/voxpupuli/splunk/util.rb
jorhett: IMHO things this deep shouldn't rescue exceptions because they're too deep to know what to do about them. That should be done higher up. IMHO only rescue exceptions here if you watch to toss your own customized exception like `Spunk::Util::DecodeException` blah blah blah IMHO YMMV

spec/acceptance/splunk_enterprise_spec.rb
```diff
@@ -1,14 +1,22 @@
...
+
   context 'default parameters' do
     # Using puppet_apply as a helper
     it 'works idempotently with no errors' do
       pp = <<-EOS
       class { '::splunk::enterprise': }
       EOS
 
-      # Run it twice and test for idempotency
+      # Run it three times and test for eventual idempotency (Needed for SELINUX limitations)
```
jorhett: can you be more specific about the need, or link to something at least for the poor lass/lad who has no idea what limitation you're referring to?

alexjfisher:
> Quick question: would it be possible for Puppet to encrypt the values in a way that Splunk finds acceptable, so as to be more declarative / less subject to changes in Splunk?

Quick question, long answer. :)

TL:DR
I've spent some time thinking about this, but think it'd actually end up being quite complicated.  The algorithm itself isn't especially tricky, but the implementation could be,

Via a function...
This forces you to manage the content of the splunks.secret file (you wouldn't want to expose it as a fact.)  You also need to decide how to generate the Initialisation Vector (16 bytes of data).  AFAIK, it doesn't *need* to be random, but it does have to be unique per use.  At the same time, you'd want the function to output the same value every run (*maybe* not required if also incorporating this change?).  You could combine fqdn_rand with a seed that users supplied and put the onus on them for it being different for each value being encrypted.

In the type...
This could be better as you can generate a random IV on the fly and use the splunks.secret file that splunk generates.  But the secrets file doesn't exist until the service is started, so it won't be available on the first run.  If it's not there, you'd have to write the values unencrypted and by the second run, splunk will have already performed the encryption itself.

I'm not actually sure either solution would be less subject to changes in Splunk.  Perhaps even the reverse.  For instance, you'd still have to make very sure you were using the correct algorithm based on installed splunk version. (There are currently 2. See https://github.com/HurricaneLabs/splunksecrets).

With all that said, if anyone wants a go at implementing a function or some munging in the type, I'd be interested in seeing what they came up with.  PRs welcome and all that! :)

spec/acceptance/splunk_enterprise_spec.rb
alexjfisher: Very fair point.  I've removed this change as it was something I only really needed when testing locally with vagrant_libvirt and it's not otherwise related to this PR.

As for the actual issue.  Puppet updates the selinux attributes on the config files on the second run since it defaults these file [properties](https://puppet.com/docs/puppet/5.4/types/file.html#file-attribute-seluser) to

> the value returned by matchpathcon for the file, if any exists

Puppet doesn't reread these default selinux contexts at any time during a run. Typically, they can be updated by a package being installed (eg splunk).  See https://tickets.puppetlabs.com/browse/PUP-2169

lib/puppet_x/voxpupuli/splunk/util.rb
alexjfisher: The thought was if decryption did fail, it could be because of something like the splunk.secrets file being replaced.  If decryption does fail (for whatever reason), it's maybe better for `insync?` to return false instead of a hard error being thrown and not handled.

I like your advise on rethrowing a custom exception and catching it higher up.  I'll leave this as is in this PR, but might work on an enhancement in a separate PR later.  Thanks for the tip.

spec/acceptance/splunk_enterprise_spec.rb
alexjfisher: Also see https://github.com/voxpupuli/puppet-splunk/pull/153#discussion_r143781103

PR #224, opened 2019-04-10
Refactor defined type Splunk::Addon

ody:
Transitions the defined type Splunk::Addon to using the splunk*_input types instead of concat, supports installation from splunkbase provided archive, and valid for use on more than just forwarder nodes.


bastelfreak:
Thanks!

manifests/addon.pp
```diff
@@ -1,72 +1,115 @@
...
-  file { "${_splunk_home}/etc/apps/${name}/local":
-    ensure => directory,
-  }
+  file { "${_splunk_home}/etc/apps/${name}/local": ensure => directory }
 
-  if $inputs {
-    concat { "splunk::addon::inputs_${name}":
-      path    => "${_splunk_home}/etc/apps/${name}/local/inputs.conf",
-      require => File["${_splunk_home}/etc/apps/${name}/local"],
+  unless $inputs.empty {
```
alexjfisher: Why? Iterating an empty hash is fine.

manifests/addon.pp
```diff
@@ -1,72 +1,115 @@
...
 
   if $splunk_home {
     $_splunk_home = $splunk_home
   }
   else {
-    $_splunk_home = $splunk::params::forwarder_homedir
+    case $mode {
+      'forwarder':  { $_splunk_home = $splunk::params::forwarder_homedir }
+      'enterprise': { $_splunk_home = $splunk::params::enterprise_homedir }
+      default:      { fail('Instances of Splunk::Addon require the declaration of one of either Class[splunk::enterprise] or Class[splunk::forwarder]') }
```
alexjfisher: This is unreachable?

manifests/addon.pp
```diff
@@ -1,72 +1,115 @@
...
   Optional[String[1]] $package_name           = undef,
   Hash $inputs                                = {},
 ) {
 
-  include 'splunk::params'
+  include splunk::params
+
+  if defined(Class['splunk::forwarder']) {
+    $mode = 'forwarder'
+  } else {
```
alexjfisher: Is it worth checking that `Class['splunk']` has been defined?

manifests/addon.pp
```diff
@@ -1,72 +1,115 @@
...
 define splunk::addon (
   Optional[Stdlib::Absolutepath] $splunk_home = undef,
   Boolean $package_manage                     = true,
+  Optional[String[1]] $splunkbase_source      = undef,
   Optional[String[1]] $package_name           = undef,
   Hash $inputs                                = {},
 ) {
 
-  include 'splunk::params'
+  include splunk::params
```
alexjfisher: Shouldn’t be necessary if it’s a requirement that one of `splunk` or `splunk:forwarder` has been included already?

ody: Yes, that should be correct. Didn't dawn me at the time.

manifests/addon.pp
ody: Class['splunk'] is now completely empty since the 7.2 refactor: https://github.com/voxpupuli/puppet-splunk/pull/215/files#diff-60ae41fd0a31977447947f59940ee9a4

manifests/addon.pp
ody: Correct and the reason I didn't originally add the default statement. Only added it to get it to pass lint checks. Not a bad thing to have, I suppose...? 

If the manifest parses this far then something strange is happening an it'd be good to explicitly fail before damage is done.

manifests/addon.pp
ody: You are correct. Just didn't realize. Tested just now and works fine without testing first

manifests/addon.pp
alexjfisher: Sorry, `splunk::enterprise` then.

ody: Wouldn't hurt.

PR #225, opened 2019-04-11
Address post-merge comments on PR #224

ody:
Removes declaration of Class[splunk::params] since it should already be in scope and declared by either Class[splunk::enterprise] or Class[splunk::forwarder] and removes an unnecessary test of an empty hash. Modifies Splunk::Addon spec test to address the removal of Class[splunk::params].

### DIFF
--- a/manifests/addon.pp
+++ b/manifests/addon.pp
@@ -43,18 +43,18 @@ define splunk::addon (
   Hash $inputs                                = {},
 ) {
 
-  include splunk::params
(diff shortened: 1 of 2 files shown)


bastelfreak:
@ody can you take a look at the failing travis jobs?

ody:
@bastelfreak That should do it, rubocop passed locally thins time...had forgot to run it beforehand

PR #227, opened 2019-04-16
Documentation style updates

ody:
Commits updates or adds "puppet strings" documentation that adheres to style guide and semantics understood by YARD. Does not prioritize the correctness or verifying the validity of documentation...

Close examination of documentation validity to be handled after manifests are consistently documented.

### DIFF
--- a/manifests/addon.pp
+++ b/manifests/addon.pp
@@ -1,4 +1,6 @@
-# @summary Defined type for deploying Splunk Add-ons and Apps from either OS packages or via splunkbase compatible archives
+# @summary
+#   Defined type for deploying Splunk Add-ons and Apps from either OS packages
+#   or via splunkbase compatible archives
 #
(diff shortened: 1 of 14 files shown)


Rakefile
```diff
@@ -4,6 +4,7 @@ require 'puppetlabs_spec_helper/rake_tasks'
 # only available if gem group releases is installed
 begin
   require 'voxpupuli/release/rake_tasks'
+  require 'puppet-strings/tasks'
```
dhoppe: I am not sure if this is necessary, because this module already has an task to create the REFERENCE.md:
* https://github.com/voxpupuli/puppet-splunk/blob/862c0b312f7e2b9521a018ca8ac0454af1aed264/Rakefile#L66

ody: @dhoppe You were correct, removed it

PR #228, opened 2019-04-19
Optionally manage splunk user and group

### LINKED ISSUE #223: Optionally manage splunk user/group
**Bug**: Changing the `splunk` user from the default (`root` on Linux) to another user will result in failures when the config files are ensured.

To reproduce, set `splunk::params::splunk_user: 'splunk' #or any group/user for this matter`

**Fix**: We should optionally manage the `splunk` user/group and default management to false:

`params.pp`:
```
Boolean $manage_splunk_user                = false,
```
`install.pp`:
```  
if $splunk::enterprise::manage_splunk_user {
    group { $splunk::enterprise::splunk_user:
      ensure  => 'present',
      require =>  Package[$splunk::enterprise::enterprise_package_name],
    }
    user { $splunk::enterprise::splunk_user:
      ensure  => 'present',
      require =>  Group[$splunk::enterprise::splunk_user],
    }
  }
```

I can have a PR staged in short order

nick-markowski:
* Declare the splunk user and group to properly ensure config files
  when the splunk user is anything but root
* Update acceptance clean-up for systemd-based installations

Fixes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,6 +369,10 @@ module. On Windows, this can be a UNC path to the MSI.  Defaults to undef.
 This variable is passed to the package resources' *install_options* parameter.
 Defaults to the value in ::splunk::params.
 
+#### `manage_splunk_user`
+
+Whether or not to manage the user splunk runuser.  Defaults to false.
+
 #### `splunk_user`
 
 The user to run Splunk as. Defaults to the value set in splunk::params.
@@ -572,6 +576,10 @@ module. On Windows, this can be a UNC path to the MSI.  Defaults to undef.
 
 This variable is passed to the package resources' *install_options* parameter.
 Defaults to the value in ::splunk::params.
+
+#### `manage_splunk_user`
+
+Whether or not to manage the user splunk runuser.  Defaults to false.
 
 #### `splunk_user`
 
--- a/manifests/forwarder/install.pp
+++ b/manifests/forwarder/install.pp
@@ -64,4 +64,16 @@ class splunk::forwarder::install {
     install_options => $splunk::forwarder::install_options,
   }
 
+  if $splunk::forwarder::manage_splunk_user {
+    group { $splunk::forwarder::splunk_user:
+      ensure  => 'present',
+      require => Package[$splunk::forwarder::forwarder_package_name],
+    }
+    user { $splunk::forwarder::splunk_user:
+      ensure  => 'present',
+      home    => $splunk::forwarder::forwarder_homedir,
+      require => Group[$splunk::forwarder::splunk_user],
+    }
+  }
+
 }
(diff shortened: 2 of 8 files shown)


nick-markowski:
I took a stab at touching up the logic to ensure users can gracefully switch the splunk user on the fly, but due to some permissions issues and having to re-set the SPLUNK_OS_USER via re-enabling boot-start, it's better suited for another commit, if at all.

manifests/forwarder/install.pp
```diff
@@ -64,4 +64,16 @@
...
 
+  if $splunk::forwarder::manage_splunk_user {
+    group { $splunk::forwarder::splunk_user:
+      ensure  => 'present',
+      require => Package[$splunk::forwarder::forwarder_package_name],
+    }
+    user { $splunk::forwarder::splunk_user:
+      ensure  => 'present',
+      home    => $splunk::forwarder::forwarder_homedir,
+      require => Group[$splunk::forwarder::splunk_user],
```
rbclark: Is there a specific reason in this ensures that the splunk package is installed first? I have a use case where I'd prefer to be able to create the splunk user and setup the filesystem under splunk with the proper permissions before installing splunk. For my use case I would have to set manage_splunk_user to false and then manually manage it instead of just taking advantage of this feature. I guess this would cause problems with the home directory not existing?

README.md
```diff
@@ -369,6 +369,10 @@ module. On Windows, this can be a UNC path to the MSI.  Defaults to undef.
 This variable is passed to the package resources' *install_options* parameter.
 Defaults to the value in ::splunk::params.
 
+#### `manage_splunk_user`
+
+Whether or not to manage the user splunk runuser.  Defaults to false.
```
ody: Is the extra "user" before "splunk" a mistake? The phrase reads oddly and you didn't type it this way in the actual parameter documentation in the manifest

README.md
```diff
@@ -573,6 +577,10 @@ module. On Windows, this can be a UNC path to the MSI.  Defaults to undef.
...
```
ody: ditto

manifests/forwarder/install.pp
ody: Quick pass through some of the various code paths in the module and it looks as if runuser is only utilized when the service starts, not at any point during the installation process so this at minimum needs to happen before Class[splunk::{enterprise, forwarder}::service] but should be fine happening after package installation...

@nick-markowski Correct me if I missed something.

bastelfreak:
Hi @nick-markowski, are you still interested in this PR?

nick-markowski:
@bastelfreak I can't replicate the issue this was supposed to fix, unless others see a need, we can close it out.

PR #233, opened 2019-04-29
fix osfamily = 'windows' and kernel = 'windows' in lowercase

kreeuwijk:
The value for kernel, osfamily and os.family in Puppet is 'windows' for the Windows OS, not 'Windows' with a capital W.


dhoppe:
@kreeuwijk Thank you very much for your contribution.

PR #234, opened 2019-04-29
Windows fixes

kreeuwijk:
This PR provides several fixes to ensure splunk::enterprise and splunk::forwarder works correctly for Windows and Linux.
It also ensures that the prerequisite net-tools package is not only ensured when needed, but that this is done before Splunk Enterprise/Forwarder is installed.


manifests/forwarder.pp
```diff
@@ -182,7 +182,7 @@
   Optional[String[1]] $package_provider      = $splunk::params::package_provider,
   Boolean $manage_package_source             = true,
   Optional[String[1]] $package_source        = undef,
-  Array[String[1]] $install_options          = $splunk::params::forwarder_install_options,
+  $install_options                           = $splunk::params::forwarder_install_options, # Not strongly typed because can be Hash or Array
```
alexjfisher: Can we not use `Variant[Hash, Array[String[1]]]`?

kreeuwijk: This is the error I get when I try that:
parameter 'install_options' variant 0 expects a Hash value, got Tuple
parameter 'install_options' variant 1 index 9 expects a String value, got Struct

manifests/forwarder.pp
```diff
@@ -172,48 +172,52 @@
...
+  String[1] $version                              = $splunk::params::version,
+  String[1] $package_name                         = $splunk::params::forwarder_package_name,
+  String[1] $package_ensure                       = $splunk::params::forwarder_package_ensure,
+  String[1] $staging_dir                          = $splunk::params::staging_dir,
+  String[1] $path_delimiter                       = $splunk::params::path_delimiter,
+  String[1] $forwarder_package_src                = $splunk::params::forwarder_package_src,
+  Optional[String[1]] $package_provider           = $splunk::params::package_provider,
+  Boolean $manage_package_source                  = true,
+  Optional[String[1]] $package_source             = undef,
+  Variant[Hash, Array[String[1]]]$install_options = $splunk::params::forwarder_install_options, # Not strongly typed because can be Hash or Array
```
alexjfisher: End of line comment no longer applicable.

manifests/forwarder.pp
kreeuwijk: Trying with `Variant[Tuple[Hash,String,1,default], Array[String[1]]]` for a reverse-order of the $install_options

manifests/forwarder.pp
kreeuwijk: fixed in commit d8ddfcd

manifests/forwarder.pp
kreeuwijk: Verified that this works.

manifests/forwarder.pp
```diff
@@ -172,41 +172,41 @@
...
+  String[1] $version                         = $splunk::params::version,
+  String[1] $package_name                    = $splunk::params::forwarder_package_name,
+  String[1] $package_ensure                  = $splunk::params::forwarder_package_ensure,
+  String[1] $staging_dir                     = $splunk::params::staging_dir,
+  String[1] $path_delimiter                  = $splunk::params::path_delimiter,
+  String[1] $forwarder_package_src           = $splunk::params::forwarder_package_src,
+  Optional[String[1]] $package_provider      = $splunk::params::package_provider,
+  Boolean $manage_package_source             = true,
+  Optional[String[1]] $package_source        = undef,
+  Variant[Tuple[Hash,String,1,default], Array[String[1]]]$install_options = $splunk::params::forwarder_install_options,
```
alexjfisher: ooh.  Since this is turning into quite a handful, it might make sense to create a custom type alias, (eg `Splunk::ForwarderInstallOptions`) ?  These can be tested in isolation. https://github.com/rodjek/rspec-puppet#type-alias-matchers

kreeuwijk: Quite possibly, but that's above my skill level unfortunately.

alexjfisher: @kreeuwijk No worries.  We can always fix that up later.  I'm going to mark this WIP so the PR doesn't accidentally get merged before you're happy you're finished.  Just remove the WIP from the PR title when you're done.

kreeuwijk: @alexjfisher ready for merging.

.puppet-lint.rc
```diff
@@ -0,0 +1,5 @@
+--no-documentation-check
+--no-140chars-check
+--no-autoloader_layout-check
+--no-class_inherits_from_params_class-check
+--no-only_variable_string
```
alexjfisher: Can you remove this file?

manifests/params.pp
```diff
@@ -204,15 +211,12 @@
...
-      $enterprise_install_options = [
-        'LAUNCHSPLUNK=1',
-        'WINEVENTLOG_APP_ENABLE=1',
-        'WINEVENTLOG_SEC_ENABLE=1',
-        'WINEVENTLOG_SYS_ENABLE=1',
-        'WINEVENTLOG_FWD_ENABLE=1',
-        'WINEVENTLOG_SET_ENABLE=1',
+      $enterprise_install_options     = [
+        { 'INSTALLDIR' => $enterprise_homedir },
+        { 'SPLUNKD_PORT' => "${splunkd_port}" },
```
alexjfisher: If you need to disable a puppet-lint warning, you can use an end-of-line [control comment](http://puppet-lint.com/controlcomments/).
Do you need to do this to convert an Integer to a string?

.puppet-lint.rc
kreeuwijk: Can you disable the `only_variable_string` check? I needed to convert and integer into a string and it triggers that warning. See [DOCUMENT-405](https://tickets.puppetlabs.com/browse/DOCUMENT-405) for why this shouldn't be triggered.

manifests/params.pp
alexjfisher: Instead of the control comment, I think you could also do...
```puppet
{ 'SPLUNKD_PORT' => String($splunkd_port) },
```

kreeuwijk: oh cool, didn't know about the control comment. Thanks for the tip!
I'll use the String() function and push the update

.puppet-lint.rc
kreeuwijk: file removed.

manifests/enterprise.pp
```diff
@@ -208,7 +208,7 @@
   Optional[String[1]] $package_provider      = $splunk::params::package_provider,
   Boolean $manage_package_source             = true,
   Optional[String[1]] $package_source        = undef,
-  Array[String[1]] $install_options          = $splunk::params::enterprise_install_options,
+  Variant[Tuple[Hash,Hash,String,3,default], Array[String[1]]] $install_options = $splunk::params::enterprise_install_options,
```
bastelfreak: can you please align the `=`? The structure in the Hashes and the string length could also be enforced, should this be a custom datatype?

manifests/enterprise.pp
```diff
@@ -250,6 +250,10 @@
     )
   }
 
+  if ($::osfamily == 'windows') and ($package_ensure == 'latest') {
```
bastelfreak: Please use the new facts hash instead:
```suggestion
  if ($facts['os']['family'] == 'windows') and ($package_ensure == 'latest') {
```

manifests/enterprise.pp
kreeuwijk: I'll align the `=` in the next commit.

The Tuple layout is dependent on the configured `$install_options`, which means that it will need updating if the `$install_options` change. For example, if another `{ 'option' => 'value' }` gets added, the Tuple structure must be updated too. I wish this wasn't as rigorous but that's how it is.

On turning it into a custom datatype, please provide the code for that as I don't know how that should be done.

bastelfreak: okay, tuple doesn't need to be more complex. You can find an example for custom datatypes in our zabbix repo.
Types are defined in a `/types/` subdirectory:
https://github.com/voxpupuli/puppet-zabbix/blob/master/types/databases.pp
and used like this:
https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/server.pp#L296

That makes them reuseable but also shorter to write. so the enterprise.pp would be easier to read.

kreeuwijk: OK, adding types in next commit

PR #241, opened 2019-06-12
Set user and group for addon installation

ody:
Commit sets user and group of expanded splunkbase archive and local directory to same as what's used for splunk installation


bastelfreak:
Thanks for the PR @ody !

PR #242, opened 2019-06-21
Setup admin account through user-seed.conf

### LINKED ISSUE #226: Setting up splunk admin could be easier
For the refactored master branch, I was getting “No users exist. Please set up a user.” in the Splunk web console after installation (using the `splunk::enterprise` class).
To solve this, I had to add the `manage_password => true` and `password_content => <valid content>` attributes.

It works with a correct password_content value but that is difficult to generate if you don’t have an existing server to glen the /opt/splunk/etc/passwd file from...

It would probably be better if one of the methods documented [here](https://docs.splunk.com/Documentation/Splunk/7.1.0/Security/Secureyouradminaccount) were used to seed a default password, so at least the admin account gets setup.

The `user-seed.conf` method didn’t work for me however. The create admin credentials via REST method did work.

ody:
Currently, the only way to ensure an admin password is set when installing Splunk is to manage the /opt/{splunk, splunkforwarder}/etc/passwd file. The issue with this approach is that when Splunk restarts it makes modifications to this file, resulting in a corrective change event every time Puppet runs on a system. The changes set forth in this PR provides a new mechanism that is more inline with recommendations from Splunk's documentation be seeding the password to the admin user at installation through user-seed.conf.

The downside of this method is that the password is set fro the admin user **ONLY** if the passwd file is non-existent upon restart of the Splunkd service. This means the password can not be continuously managed through Puppet without triggering a corrective change event in this way either so management only occurs at initial installation and optionally, in a temporary fashion if someone explicitly sets the appropriate variable on either Class[splunk::enterprise] or Class[splunk::forwarder].

I've chosen to use new parameters instead of repurposing the old ones to signify the change of behavior.

Fixes #226


ody:
I've validated it works but haven't added documentation yet or looked at if their are any valuable tests to add

nick-markowski:
@ody do you think it's worth trying to break out the password management to a password class so users could manually puppet apply and restart splunk without needing to apply the entire enterprise or forwarder class?  I believe we could preserve existing functionality if we instantiated the password class in config.pp

ody:
Great idea Nick, makes it much more Bolt friendly for doing the one off password changes. I am on an airplane and out for the next 10 days on PTO so won’t be able to work on it until the 10th.

ody:
@nick-markowski Password management moved into its own classes. Kept implementations between seeding, direct management, and enterprise vs. forwarder separate to keep this consistent with rest of the module.

Gemfile
```diff
@@ -59,6 +59,7 @@ group :system_tests do
   gem 'rbnacl', '>= 4',                     :require => false
   gem 'rbnacl-libsodium',                   :require => false
   gem 'bcrypt_pbkdf',                       :require => false
+  gem 'ed25519',                       :require => false
```
alexjfisher: This file is managed by modulesync, so you'll need to add this gem to `.sync.yml` (and could you also fix the indentation).
See https://github.com/voxpupuli/puppet-redis/blob/d0ee587d0d0327ef85d5701fbd6c9d9f33569fbe/.sync.yml#L10-L14 for an example.

alexjfisher:
Do you want to add one or two lines to the README?

Gemfile
ody: Thanks, slipped it while doing local testing. I can't easily run acceptance tests for any module locally without that gem so I'll send a PR to add to: https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml

ody:
@alexjfisher Sure. Side note though: That README needs a re-write.

alexjfisher:
@ody Thanks!

PR #245, opened 2019-07-19
Add savedsearches.conf to managed files

### LINKED ISSUE #244: Ability to manage savedsearches.conf
Would be nice to be able to add alerts to new servers programmatically. Should just be a copy of the existing splunk_* providers/types to splunk_savedsearches:

bsg-sfrazer:
Adds the "savedsearches.conf" file to the list of managable files 

Fixes #244


metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-splunk",
-  "version": "7.3.1-rc0",
+  "version": "7.3.2",
```
bastelfreak: Please don't update the version. We will handle this during the release process

bastelfreak:
Hi @bsg-sfrazer, thanks for the PR. Can you please rebase against our master branch and squash the commits into one? Can you also add tests for the new parameters?

bsg-sfrazer:
Hi Tim,

Sorry about the release version, we needed it so librarian would update.

We actually stopped using this functionality fairly quickly. Turns out: if a user saves a search through the web interface, this will overwrite it

I’ve been working on a way to distribute the searches through Splunk Apps using this module, but I’ve run into a versioning issue: if you supply an app tarball it will install fine but there’s no way to get it to update the tarball should anything change.

I’m playing with adding a versioning string to the Splunk::addons class

________________________________

PR #250, opened 2019-07-31
WIP: Fix issues 230, 232, 235, and 249

### LINKED ISSUE #230: Can you please let me know when will be this update reflect in the puppet forge module?
Can you please let me know when will be this update reflect in the puppet forge module?

### LINKED ISSUE #232: `splunk::enterprise` version parameter doesn't work
See https://github.com/voxpupuli/puppet-splunk/issues/230#issuecomment-487060813

### LINKED ISSUE #235: Duplicate declaration: Class[Splunk::Params] is already declared
Hi,

I test the current version of this module and had the following issue:
` Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Splunk::Params] is already declared`

My Puppet profile looks like this:
```
class profile::splunk {

  include splunk::enterprise
  
  class { 'splunk::prams':
    enterprise_src_root => 'http://...'
  }
}
```

Did i missunderstand the module documentation or i am on the wrong way? In my understanding the issue comes with the inheritance of the params class, which will be automatic defined by the class splunk::enterprise.

How can I provide my own values for the splunk::enterprise module? I think it makes no sense to make the params class parameterized.

regards,
Martin

### LINKED ISSUE #249: Add a $service_ensure param to splunk::enterprise and splunk::forwarder
Service ensure defaults to 'running' for the Splunk Enterprise and Splunk Forwarder services. It should be a parameter so users can toggle it.

Issue created per: https://github.com/voxpupuli/puppet-splunk/pull/243

nick-markowski:
* Moved defaults to module data, and removed the params class
* Privatized all `splunk::enterprise::` and `splunk::forwarder::` install, config, and service classes
* Added a `$release` param, which replaces the `$version` param
  * For ensurable package_providers, the release is used as the Splunk package ensure, if specified
  * The release no longer defaults to a specific version and build, instead, the Splunk package resource defaults ensure to 'installed'
  * Added a Splunk::Release type
* Added a service_ensure param, per #249
* Modified splunk*_version facts to be part of splunkforwarder and splunkenterprise fact hashes
* Removed init.pp, which only served to confuse
* `$[enterprise,forwarder]_package_src` and `$package_source` params renamed to `$managed_package_source` and `$unmanaged_package_source`, for clarity
* Fixed: enterprise and forwarder password classes cross-referenced params

Fixed #230
Fixed #232
Fixed #235
Fixed #249


nick-markowski:
The bulk of the changes are done, but here are a few notes about the initial commit:
* This commit finalizes what the original refactor should have been.  The module much more closely follows the module best practices, and the refactor was necessary to fix a number of the issues reported by users.
* Still need to expand upon the README and update REFERENCES.md.
* I've been experiencing an issue with splunk_config redefining types and throwing errors during puppet-applys, for instance:
```
  Info: Redefining splunk_distsearch in Puppet::Type
  Error: Could not autoload puppet/type/splunk_config: Attempt to redefine entity 'http://puppet.com/2016.1/runtime/type/splunk_distsearch'. Originally set at file:/etc/puppetlabs/code/modules/splunk/lib/puppet/type/splunk_distsearch.rb?line=3.
  Error: Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/splunk_config: Attempt to redefine entity 'http://puppet.com/2016.1/runtime/type/splunk_distsearch'. Originally set at file:/etc/puppetlabs/code/modules/splunk/lib/puppet/type/splunk_distsearch.rb?line=3. (file: /etc/puppetlabs/code/modules/splunk/manifests/enterprise.pp, line: 311, column: 3) on node centos7-64-1.home
```

manifests/enterprise/password/manage.pp
```diff
@@ -21,27 +21,31 @@
...
 #
 class splunk::enterprise::password::manage(
-  Boolean $manage_password                   = $splunk::params::manage_password,
-  Stdlib::Absolutepath $password_config_file = $splunk::params::forwarder_password_config_file,
-  String[1] $password_content                = $splunk::params::password_content,
-  Stdlib::Absolutepath $secret_file          = $splunk::params::forwarder_secret_file,
-  String[1] $secret                          = $splunk::params::secret,
-  String[1] $splunk_user                     = $splunk::params::splunk_user,
-  String[1] $service                         = $splunk::params::enterprise_service,
+  Boolean $manage_password                   = lookup($splunk::enterprise::manage_password),
```
bastelfreak: Why did you add all the `lookup()` calls?

nick-markowski: After thinking about it, that was silly, and I'm going to remove them

data/common.yaml
```diff
@@ -0,0 +1,57 @@
+---
+splunk::enterprise::manage_password: false
+splunk::enterprise::seed_password: false
+splunk::enterprise::reset_seeded_password: false
+splunk::enterprise::secret: 'hhy9DOGqli4.aZWCuGvz8stcqT2/OSJUZuyWHKc4wnJtQ6IZu2bfjeElgYmGHN9RWIT3zs5hRJcX1wGerpMNObWhFue78jZMALs3c3Mzc6CzM98/yGYdfcvWMo1HRdKn82LVeBJI5dNznlZWfzg6xdywWbeUVQZcOZtODi10hdxSJ4I3wmCv0nmkSWMVOEKHxti6QLgjfuj/MOoh8.2pM0/CqF5u6ORAzqFZ8Qf3c27uVEahy7ShxSv2K4K41z'
+splunk::enterprise::password_hash: '$6$pIE/xAyP9mvBaewv$4GYFxC0SqonT6/x8qGcZXVCRLUVKODj9drDjdu/JJQ/Iw0Gg.aTkFzCjNAbaK4zcCHbphFz1g1HK18Z2bI92M0'
+splunk::enterprise::password_content: ":admin:${password_hash}::Administrator:admin:changeme@example.com::"
```
ody: Does embedding variables in this way work hiera the same as it did with manifest interpolation? Documentation seems to indicate, no.

https://puppet.com/docs/puppet/latest/hiera_merging.html#the-lookup-and-hiera-function

ody: Corroborated by the fact that this was caught in other data files, e.g. Linux.yaml

lib/facter/splunkenterprise.rb
```diff
@@ -1,17 +1,18 @@
-Facter.add(:splunk_version) do
+Facter.add(:splunkenterprise) do
   setcode do
-    value = nil
+    splunk_hash = {}
```
ody: I've never encountered another fact that returns an empty hash like this. The behavior I'd expect is that if there is no data related to splunk then the fact should be undefined within a manifest. If the behavior isn't followed then this fact will exist on all nodes within an infrastructure due to the way pluginsync works.

ody (changes requested):
Overall looks really good. Nice work!

nick-markowski:
@ody Thank you for the review! I've *really* dragged my feet digging into this PR after submitting.  By now it'll probably need a nasty rebase, but I'll try to get the feedback addressed and get things rolling again.

data/common.yaml
nick-markowski: Yeah, this should certainly be a lookup

lib/facter/splunkenterprise.rb
nick-markowski: Agreed, I'll make some changes

PR #252, opened 2019-08-02
Make sure package is installed before calling execs

coreone:
To correct ordering in this module for our installations, we have had to do the following:

```puppet
Package[$::splunk::forwarder::forwarder_package_name] -> Exec[stop_splunkforwarder]
```

This seemed to just be a missing `require` in the module, which this PR attempts to correct.

n/a


coreone:
@ghoneycutt I missed the tag on this PR.  I added tests for the exec changes I made, but it looks as if the spec tests are failing on master now for `solaris-11-i86pc`.

alexjfisher:
Was the ordering fixed in https://github.com/voxpupuli/puppet-splunk/pull/265?

coreone:
#265 did indeed fix this issue, so I will close this out now.  Thanks!

PR #253, opened 2019-08-06
 refactored splunkforwarder_version fact 

posteingang:


bastelfreak:
@posteingang thanks for the PR. Can you take a look at the failing spec tests?

PR #255, opened 2019-08-27
make sure service::nix is completed before starting the service

ashish1099:
With `include` it was failing. So had to move it to `contain` which will make sure that its setup the licences thing.

```
Error: Could not start Service[splunk]: Execution of '/sbin/service splunk start' returned 2: Starting Splunk...
License not yet accepted, but executed with no-prompt flag.  Exiting.
Error: /Stage[main]/Splunk::Forwarder::Service/Service[splunk]/ensure: change from 'stopped' to 'running' failed: Could not start Service[splunk]: Execution of '/sbin/service splunk start' returned 2: Starting Splunk...
License not yet accepted, but executed with no-prompt flag.  Exiting. (corrective)
```


ashish1099:
It was failing because of the stale init.d/splunk file

PR #260, opened 2019-10-04
Fix flaky acceptance tests

ody:
This commit adds relationship that's getting lost in a sea of
	containment layers that ensures Splunk services are restarted
	before the test checks for success or not.

	Without this the test fails in certain random situations and
	likely in a limited number of real world cases as well.


bastelfreak:
thanks for the bugfix!

PR #267, opened 2019-11-13
drop Ubuntu 14.04 support

bastelfreak:


bastelfreak:
I also looked for 14.04 specific code but didn't find anything related to that.

PR #276, opened 2020-02-19
adding config option for https

jose-lmax:


manifests/forwarder/password/seed.pp
```diff
@@ -57,7 +57,7 @@
     content => $secret,
   }
 
-  if $reset_seeded_password or $facts['splunk_version'].empty {
+  if $reset_seeded_password or $facts['splunkforwarder_version'].empty {
```
bastelfreak, line 60: this looks like a bugfix that isn't related to the other change? If so, can you please submit it as a separate pull request?

jose-lmax: will do

PR #278, opened 2020-02-24
Check if splunk forwarder was already installed when seeding password.

mattk42:
Currently when seeding passwords for splunk::forwarder, it will continually re-seed unless you also have the splunk enterprise installed. This should resolve that.

n/a


bastelfreak:
thanks for the fix!

PR #279, opened 2020-03-03
Add ability to override default-host configuration

dave-pollock:
Add ability to override default-host configuration. The default-host configuration is used to control the "host" property which Splunk will use to identify the forwarder. The current behaviour forces the default-host to match the server's $HOSTNAME. In my environment we prefer to customise this name to identify the region and environment for the forwarder (info we don't normally include in the server's hostname).

The default behaviour is maintained so this shouldn't break any existing setup. It just adds the ability to override it if required.

N/A

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -106,7 +106,7 @@ class splunk::params (
     'windows' => 'Administrator',
     default => 'root'
   },
-  Optional[String[1]] $default_host          = $facts['clientcert'],
+  String[1] $default_host                    = $facts['clientcert'],
 ) {
   # Based on the small number of inputs above, we can construct sane defaults
   # for pretty much everything else.
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -103,6 +103,7 @@ class splunk::params (
     'windows' => 'Administrator',
     default => 'root'
   },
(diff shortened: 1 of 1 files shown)


dave-pollock:
I don't really see how these test failures could be related to my change?

Please let me know if I'm missing something. Cheers!

dave-pollock:
@bastelfreak since you seem to be the most active maintainer would you mind reviewing this PR if you have a minute free? Thanks.

dave-pollock:
@ghoneycutt perhaps you could take a look at this if you have a moment?

bastelfreak:
@dave-pollock can you please rebase this against our latest master? The git history is a bit clumsy at the moment.

dave-pollock:
@bastelfreak I've rebased.

manifests/params.pp
```diff
@@ -103,6 +106,7 @@
     'windows' => 'Administrator',
     default => 'root'
   },
+  Optional[String[1]] $default_host          = $facts['clientcert'],
```
bastelfreak, line 109: since the default isn't undef, you cannot pass undef to it. This means we can default to String as datatype:
```suggestion
  String[1] $default_host                    = $facts['clientcert'],
```

PR #281, opened 2020-03-23
Add support for OpenSUSE

martijndegouw:
Add support for OpenSUSE. Tested on OpenSUSE 15


bastelfreak:
@martijndegouw thanks again for the PR. Please also take a look at the failing tests and add OpenSUSE to the metadata.json

martijndegouw:
I looked into the failing tests, but it looks like the problem is in the test, now in the newly added code:
Test is running on solaris-11-i86pc and returns unsupported osfamily/arch Solaris/ (params.pp:337)

Looks like ${facts['os']['architecture']} is not set correctly in the correspondig ruby/puppet version

bastelfreak:
thanks for the work!

PR #282, opened 2020-03-23
Add support for FreeBSD

martijndegouw:
Add support for FreeBSD. Tested on FreeBSD 12.1

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -194,6 +194,27 @@ class splunk::params (
         $supports_systemd        = false
       }
     }
+    'FreeBSD': {
+      $path_delimiter                  = '/'
+      $forwarder_src_subdir            = 'freebsd'
+      $forwarder_seed_config_file      = "${forwarder_homedir}/etc/system/local/user-seed.conf"
+      $enterprise_seed_config_file     = "${enterprise_homedir}/etc/system/local/user-seed.conf"
+      $forwarder_password_config_file  = "${forwarder_homedir}/etc/passwd"
+      $enterprise_password_config_file = "${enterprise_homedir}/etc/passwd"
+      $forwarder_secret_file           = "${forwarder_homedir}/etc/splunk.secret"
+      $enterprise_secret_file          = "${enterprise_homedir}/etc/splunk.secret"
+      $forwarder_confdir               = "${forwarder_homedir}/etc"
+      $enterprise_src_subdir           = 'freebsd'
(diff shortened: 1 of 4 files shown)


bastelfreak:
@martijndegouw thanks for the PR! Can you u take a look at the failing spec tests and also add FreeBSD to the metadata.json?

bastelfreak (approved):

bastelfreak:
@martijndegouw thanks for the updates. can you take a look at the failing tests?

spec/classes/forwarder_spec.rb
```diff
@@ -230,50 +230,54 @@
           end
 
           context 'when forwarder not already installed' do
```
bastelfreak, line 232: instead of the additional level in indentation, you can do the following:
```suggestion
          context 'when forwarder not already installed', unless: facts[:kernel] == 'FreeBSD' do
```

martijndegouw: I can change that, if that's the preferred way. I just copied it over from the tests above.

martijndegouw: I guess it would be nice to have a test for service_provider == init as well. I'll give that a try.

manifests/params.pp
```diff
@@ -261,6 +282,7 @@
     'Debian':  { $package_provider = 'dpkg' }
     'Solaris': { $package_provider = 'sun' }
     'Suse':    { $package_provider = 'rpm' }
+    'FreeBSD': { $package_provider = 'pkgng' }
```
bastelfreak, line 285: Is setting this required? in 99% of the cases undef is fine and puppet will figure it out automatically.

martijndegouw: I just added it for completeness, but it looks like there is code depending on it in https://github.com/voxpupuli/puppet-splunk/blob/99442a75bce646ae8795a620f2b0c2635eceec38/manifests/forwarder/install.pp#L12 
To be honest, I did not test it without setting is specifically.

spec/classes/forwarder_spec.rb
```diff
@@ -231,7 +231,7 @@
...
             let(:facts) do
-              facts.merge(splunkforwarder_version: nil, service_provider: 'systemd')
+              facts.merge(splunkforwarder_version: '7.3.3', service_provider: facts[:kernel] == 'FreeBSD' ? 'freebsd' : 'systemd')
```
martijndegouw, line 234: wh00ps

PR #291, opened 2020-07-14
(docs) cleaned up the TBD sections.

binford2k:
Can we also validate the Limitations for current relevance?


binford2k:
Failing check is unrelated to this PR

Issue #275, opened 2020-02-17
Hard coded dependency net-tools


jose-lmax:
Affected: Puppet

## How to reproduce (e.g Puppet code you use)
class profile::splunk_master(
  $version = latest,
  $build = undef,
  $managepassword = true,
  $splunkuser = 'splunk_admin',
  $password = 'secret',
  $webport = 443
) inherits profile {
  ensure_packages(['net-tools'])

  class { 'splunk::params':
    version => $version,
    build   => $build,
  }

  class { 'splunk::enterprise':
    package_ensure   => latest,
    manage_password  => $managepassword,
    splunk_user      => $splunkuser,
    package_provider => 'rpm',
    web_httpport     => $webport,
    password_hash    => $password,
  }

## What are you seeing
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: Package[net-tools] is already declared; cannot redeclare (file: /etc/puppetlabs/code/environments/production/modules/splunk/manifests/enterprise/install/nix.pp, line: 36, column: 5) on node splunk.lmax

## What behaviour did you expect instead
To don't get that error, the dependency of that package should not be hard coded in nix.pp, probably the solution is to use require => Package['net-tools']

PR #294, opened 2020-08-03
implement possibility to manage dependencies or not.

the-player777:
implement the manage_dependencies variable to manage external dependencies for forwarder and entreprise.

Solving issue #275


bastelfreak:
@headyj ping :)

PR #298, opened 2020-11-02
Fix rspec 3.10 failures

bastelfreak:


ekohl (approved):
There's more things we could clean up, like 
https://github.com/voxpupuli/puppet-splunk/blob/b0e709c9b76a95baa984f5e2f2440d45e03956ce/spec/classes/enterprise_spec.rb#L55

But I'm fine with just fixing the tests for now.

PR #303, opened 2020-11-25
add option to manage net-tools package

mcmartins:
I believe this is a controversial pull request.

net-tools is very common dependency, and in a huge code base with many dependencies most likely it will be already declared somewhere.

I've introduced a flag `manage_net_tools` in params.pp with a default `true`, which will keep working as it is for everyone, but will allow projects like ours to mark it as `false` when net-tools is already managed somewhere else.

Thanks.

Manuel

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -266,6 +266,18 @@ manage the Splunk Enterprise service.
 
 Default value: $splunk::params::boot_start
 
+##### `manage_net_tools`
+
+Data type: `Boolean`
+
+Whether or not to declare the resource Package[net-tools].
+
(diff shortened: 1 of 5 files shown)


kenyon:
Why does Splunk even need a package that's been obsolete/deprecated for like 20 years? 😞

mcmartins:
My bad, I'll include some tests tomorrow.

mcmartins:
Added some tests on `enterprise_spec.rb` and updated the `REFERENCE.md`

bastelfreak:
thanks for the PR. I rebased it against our latest master to pull in the updated CI matrix.

PR #309, opened 2021-04-16
puppet/archive: allow 5.x

bastelfreak:


bastelfreak:
This is expected to fail because Puppet does not provide Puppet 5 packages for Ubuntu 20

foxxx0 (approved):

PR #287, opened 2020-05-18
Read Splunk versions from text file

runejuhl:
#### Pull Request (PR) description
This is to avoid issues with `splunk --version` waiting on input if the license
has not yet been accepted. See issues
https://github.com/voxpupuli/puppet-splunk/issues/238 and
https://github.com/voxpupuli/puppet-splunk/issues/280.

The code is based on
https://github.com/posteingang/puppet-splunk/commit/09092f64e9fed3bb719d94a5cab101b0ef234775
with minor changes.

#### This Pull Request (PR) fixes the following issues
Fixes #238
Fixes #280

PR #320, opened 2021-09-15
Safer version check

mvandegarde:
This prevents facter from getting stuck in an interactive session.

### DIFF
--- a/lib/facter/splunkforwarder_version.rb
+++ b/lib/facter/splunkforwarder_version.rb
@@ -4,7 +4,7 @@ Facter.add(:splunkforwarder_version) do
     cmd = if File.exist?('C:/Program Files/SplunkUniversalForwarder/bin/splunk.exe')
             '"C:/Program Files/SplunkUniversalForwarder/bin/splunk.exe" --version'
           elsif File.exist?('/opt/splunkforwarder/bin/splunk')
-            '/opt/splunkforwarder/bin/splunk --version'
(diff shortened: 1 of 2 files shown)


smortex (changes requested):
Thanks!  Please update the test suite accordingly :wink:

mvandegarde:
> Thanks! Please update the test suite accordingly 😉

Should be ok now. Unfortunately I can't verify if this works the same on the windows binary, or on the regular splunk_binary. I'd be happy to do the same there but i can't even find the parameters i'm adding right now in the splunk forwarder documentation, i've only tested & confirmed that they do work.

smortex (commented):
Ah…  we do not have CI for windows so who knows :shrug:.  Maybe windows users will send a PR to fix windows support if it broke and we will issue a new patch release then so probably not a big issue.

Can you please combine your commits in a single one?

From your working directory:

```
git rebase -i origin/master # Interactively rewrite history
```

This will bring your editor.  Keep the first line unchanged, and replace `pick` with `squash` for all subsequent lines.  Save and quit, your editor will allow you to combine the commit messages in a single commit message.

Then push your updated branch:

```
git push -f            # Send the changes (-f is required because we re-wrote history)
```

Thanks!

smortex (approved):

root-expert:
This PR superseded 

* #287 
* #253

PR #323, opened 2021-09-20
Allow user-seed.conf to set a username besides 'admin'

gcoxmoz:
Allow `user-seed.conf` to have a not-hardcoded userid instead of 'admin.

The `user-seed.conf` file sets a default local user/password (admin/changeme by default).  Some security policies wish for you to have only non-default userids.  At present, the existing module doesn't allow that.

The testing is a little lax here because there was not great testing on these classes to begin with.  Sorry.

### DIFF
--- a/templates/user-seed.conf.epp
+++ b/templates/user-seed.conf.epp
@@ -1,4 +1,5 @@
-<%- | String $hash | -%>
+<%- | String $user,
+      String $hash | -%>
 [user_info]
-USERNAME=admin
+USERNAME=<%= $user %>
 HASHED_PASSWORD=<%= $hash %>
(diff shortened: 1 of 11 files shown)


templates/user-seed.conf.epp
```diff
@@ -1,4 +1,5 @@
-<%- | String $hash | -%>
+<%- | String $user,
+      String $hash | -%>
 [user_info]
-USERNAME=admin
+USERNAME=<%= $user %>
```
bastelfreak, line 4: thanks for the PR, can you please add a unit test for this change?

gcoxmoz: The class had no test at all, so I kinda winged it, but, "added."

bastelfreak:
thanks for the work!

PR #331, opened 2022-04-07
Add support for using a proxy to install Splunk

haught:
This PR adds support to use a given proxy for install the forwarder and enterprise Splunk. It simply passes the options to puppet-archive.

### DIFF
--- a/manifests/enterprise.pp
+++ b/manifests/enterprise.pp
@@ -171,6 +171,12 @@
 # @param secret
 #   The secret used to salt the splunk password.
 #
+# @param install_proxy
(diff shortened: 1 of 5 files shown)


bastelfreak:
thanks for the PR! Can you add some tests for the changes?

haught:
I looked through the tests before I PR'd and really didn't see anything like it that would test for this change. The change doesn't change a config file or anything, it just passes along two variables to puppet-archive. I am not even sure what I would be testing for.

kenyon:
You could test the values of the archive resource attributes given various inputs.

PR #333, opened 2022-06-23
Solaris support

### LINKED ISSUE #272: Solaris installs point to an old package, and have a dependency on net-tools

- Puppet:
- Ruby:
- Distribution:
- Module version: 7.3.0

 ```puppet
 if $facts['osfamily'] == 'Solaris' {
      class { '::splunk::params' :
        build => $solaris_build,
      }
  }

  class { '::splunk::forwarder' :
    package_ensure => 'latest',
  }
```
## What are you seeing
The package version that the module selects is `/opt/staging/splunk/splunkforwarder-7.3.1.1-8e225e1518e2-solaris-8-sparc.pkg`

The -solaris-8-sparc.pkg is set in this case statement in params.pp
```puppet
    'Solaris sun4v': {
      $package_suffix          = "${version}-${build}-solaris-8-sparc.pkg"
      $forwarder_package_name  = 'splunkforwarder'
      $enterprise_package_name = 'splunk'
    }
```
Unless I am missing it there are not any solaris-8-sparc.pkg files provided by splunk since the 6.2.6 version.

Also, for the SunOS kernel the module requires a net-tools package, which is not required for Solaris as far as I can tell.  (Feel free to correct me if I am mistaken here).

I expected the module to select the appropriate package for Solaris 10 and 11
For Solaris 10: -solaris-10-sparc.pkg.Z
For Solaris 11: -solaris-11-sparc.p5p

I also expected the net-tools requirement to not be enforced on Solaris hosts.

### LINKED ISSUE #310: Forwarder service won't start due to incorrect permissions
## Summary ##
When using splunkforwarder 8.1.3, configured to start as the root user, it fails to start.  The `splunk enable boot-start ...` command configures the systemd service to start as `splunk`, even though the Puppet module is configured to start it as `root`.

- Puppet: 6.19.1
- Distribution: RHEL7
- Module version: 8.0.0

Configure the module to run the forwarder as root.  One way to do it is by adding this to Hiera:
```yaml
splunk::forwarder::splunk_user: 'root'
```
Run Puppet on a host that doesn't have splunkforwarder installed yet.

The systemd file will be configured to run the service as the splunk user.

## What are you seeing
Puppet runs this command to enable the forwarder:
```
/opt/splunkforwarder/bin/splunk enable boot-start  -systemd-managed 1 --accept-license --answer-yes --no-prompt
```

When running the command manually, it shows this output:
```
# /opt/splunkforwarder/bin/splunk enable boot-start  -systemd-managed 1 --accept-license --answer-yes --no-prompt

This appears to be your first time running this version of Splunk.

IMPORTANT: Because an admin password was not provided, the admin user
 will not be created. You will have to set up an admin username/password
 later using user-seed.conf.
Important: splunk will start under systemd as user: splunk
Systemd unit file installed at /etc/systemd/system/SplunkForwarder.service.
Configured as systemd managed service.
```

Note that by leaving out the `-user` parameter, it configures systemd to run the service as the splunk user.  This is in contradiction to the Hiera setting where it was designated to run as root.

```console
# grep -E '(User|Group)' /etc/systemd/system/SplunkForwarder.service
User=splunk
Group=splunk
```

The module correctly sets ownership on some files in /opt/splunkforwarder to root:root, but because the service is running as the splunk user, it cannot read those files and the service fails to start.

I expect the service to run as the root user, because that's what I set via `splunk::forwarder::splunk_user`.

I suspect the issue is related to this code:
https://github.com/voxpupuli/puppet-splunk/blob/14fef3618df8faf5003eaf49dd1b5d131cb949d9/manifests/forwarder/service/nix.pp#L22-L26

Notice that if systemd is used and the user is root, it leaves out the `-user` parameter.  I suspect this worked in the past because without the parameter, the "splunk enable" command used root by default, but in 8.x or some other version they changed it to default to the splunk user instead.

Also see https://github.com/voxpupuli/puppet-splunk/issues/262 for history.

Anyway, I don't see any harm in always including the `-user` parameter, so the fix may be as easy as getting rid of that conditional.

davemcdonnell:
Fixes for permission startup when running as root user. 
Removal of net-tools due to this incorrectly being specified for SunOS.

Fixes #310 - it appears the default user for splunk has changed from root to splunk at some point. The default being root meant not explicitly specifying the user still worked. The resulted in the module config having root, file ownership having root, but the service starting as splunk due to it not being explicitly specified.

Partially fixes #272 - Net-tools removed. On the other issue reported, using the .pkg file, it appears Splunk no longer ship a .pkg format installer. Only .p5p and .Z. I had issues getting .p5p support working because I'm using Solaris zones, and this is incompatible with the way the installer works (Solaris zones need a pkg provider defined in the global zone, and cant install .p5p files without this).

My workaround for the missing .pkg download was to build a .pkg file myself using pkgbuild. Due to licensing I dont think I'm able to distribute the build package but I've put pkgbuild spec file and some instructions in this git repo:
https://github.com/davemcdonnell/splunk-solarispkg

With all of the above I've got splunk forwarder running on Solaris 11.4.45 with splunkforwarder 9.0.0.

### DIFF
--- a/manifests/forwarder/install.pp
+++ b/manifests/forwarder/install.pp
@@ -61,7 +61,7 @@ class splunk::forwarder::install {
   }
 
   # Required for splunk 7.2.4.2
-  if $splunk::params::manage_net_tools and ($facts['kernel'] == 'Linux' or $facts['kernel'] == 'SunOS') and (versioncmp($splunk::forwarder::version, '7.2.4.2') >= 0) {
+  if $splunk::params::manage_net_tools and $facts['kernel'] == 'Linux' and (versioncmp($splunk::forwarder::version, '7.2.4.2') >= 0) {
     ensure_packages(['net-tools'], {
         'ensure' => 'present',
     })
--- a/manifests/forwarder/service/nix.pp
+++ b/manifests/forwarder/service/nix.pp
@@ -23,6 +23,13 @@ class splunk::forwarder::service::nix inherits splunk::forwarder::service {
     } else {
       $user_args = "-user ${splunk::forwarder::splunk_user}"
     }
+
+    if $facts['kernel'] == 'SunOS' {
+      Service[$splunk::forwarder::service_name] {
+        provider => 'init',
+      }
+    }
+
     # This will fail if the unit file already exists.  Splunk does not remove
     # unit files during uninstallation, so you may be required to manually
     # remove existing unit files before re-installing and enabling boot-start.
--- a/manifests/forwarder/service/nix.pp
+++ b/manifests/forwarder/service/nix.pp
@@ -18,9 +18,8 @@ class splunk::forwarder::service::nix inherits splunk::forwarder::service {
       timeout => 0,
       notify  => Exec['enable_splunkforwarder'],
     }
-    if $splunk::params::supports_systemd {
-      $user_args = "-user ${splunk::forwarder::splunk_user}"
-    }
+
+    $user_args = "-user ${splunk::forwarder::splunk_user}"
 
     if $facts['kernel'] == 'SunOS' {
       Service[$splunk::forwarder::service_name] {
--- a/manifests/forwarder/service/nix.pp
+++ b/manifests/forwarder/service/nix.pp
@@ -18,9 +18,7 @@ class splunk::forwarder::service::nix inherits splunk::forwarder::service {
       timeout => 0,
       notify  => Exec['enable_splunkforwarder'],
     }
-    if $splunk::params::supports_systemd and $splunk::forwarder::splunk_user == 'root' {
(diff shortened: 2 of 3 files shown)


davemcdonnell:
Test failing, i'll review, and separate out the fixes in separate commits and make sure tests are passing

davemcdonnell:
I believe I've fixed the tests but unable to 100% verify locally due to some issues getting tests to run (the relevant test to this pull req are passing now).

I've run out of time to get the commits separated out for each individual issue. I wanted to reopen this PR to see if tests pass, and maybe someone can finish it up.

davemcdonnell:
Replaced by:
#334
#335 
#336

PR #340, opened 2023-02-28
enable the puppet-archive `allow_insecure` parameter

mwpower:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -97,6 +97,9 @@
 #   By default this module manages the resource Package[net-tools], if this resource is
 #   already declared on your code base, you can disable this flag.
 #
+# @param allow_insecure
+#   Disable certificate verification when connecting to SSL hosts to download packages.
+#
 class splunk::params (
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -97,6 +97,10 @@
 #   By default this module manages the resource Package[net-tools], if this resource is
 #   already declared on your code base, you can disable this flag.
 #
+# @param allow_insecure
+#   Disable certificate verification when connecting to SSL hosts to download packages.
+#   Defaults to false
```
kenyon, line 102: Puppet Strings docs automatically generate the defaults, so no need to write it.

```suggestion
```

kenyon (approved):
LGTM. The tests are failing because this module badly needs some maintenance, e.g., #322.

manifests/params.pp
mwpower: Thanks; submitted that change.

kenyon (approved):

PR #363, opened 2024-03-22
Fix upgrading Splunk Enterprise

siegy22:
When upgrading, there was an issue with accepting the license of the new version. Thus symptoms described in #339 happened. To ensure we have a clean upgrade we need to re-enable boot-start after the upgrade.

This will also apply the correct systemd service file when upgrading, before it was just the service file of the originally installed version. It never changed.

### DIFF
--- a/manifests/enterprise/service/nix.pp
+++ b/manifests/enterprise/service/nix.pp
@@ -16,6 +16,19 @@ class splunk::enterprise::service::nix inherits splunk::enterprise::service {
       timeout => 0,
       notify  => Exec['enable_splunk'],
     }
+
+    exec { 'disable_splunk_boot_start':
+      subscribe   => Package[$splunk::enterprise::package_name],
+      refreshonly => true,
+      command     => [
+        "${splunk::enterprise::enterprise_homedir}/bin/splunk",
+        'disable',
+        'boot-start',
+        '--accept-license',
+        '--answer-yes',
+        '--no-prompt',
+      ],
+    }
     if $splunk::params::supports_systemd and $splunk::enterprise::splunk_user == 'root' {
       $user_args = ''
     } else {
@@ -29,6 +42,7 @@ class splunk::enterprise::service::nix inherits splunk::enterprise::service {
       refreshonly => true,
       before      => Service[$splunk::enterprise::service_name],
       require     => Exec['stop_splunk'],
+      subscribe   => Exec['disable_splunk_boot_start'],
     }
   }
   # Commands to license, disable, and start Splunk Enterprise
--- a/spec/acceptance/splunk_enterprise_spec.rb
+++ b/spec/acceptance/splunk_enterprise_spec.rb
@@ -55,6 +55,70 @@ describe 'splunk enterprise class' do
     end
   end
 
+  context 'Upgrading splunk enterprise from 9.1.0 to 9.2.0.1' do
+    after(:all) do
+      pp = <<-EOS
+        service { '#{service_name}': ensure => stopped }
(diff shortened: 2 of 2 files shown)


manifests/enterprise/service/nix.pp
```diff
@@ -16,6 +16,12 @@
       timeout => 0,
       notify  => Exec['enable_splunk'],
     }
+
+    exec { 'disable_splunk_boot_start':
+      command     => "${splunk::enterprise::enterprise_homedir}/bin/splunk disable boot-start --accept-license --answer-yes --no-prompt",
```
bastelfreak, line 21: can you please provide the command as array? see https://www.puppet.com/docs/puppet/8/types/exec.html#exec-attribute-command

manifests/enterprise/service/nix.pp
```diff
@@ -28,7 +34,8 @@
       command     => "${splunk::enterprise::enterprise_homedir}/bin/splunk enable boot-start ${user_args} ${splunk::params::boot_start_args} --accept-license --answer-yes --no-prompt",
       refreshonly => true,
       before      => Service[$splunk::enterprise::service_name],
-      require     => Exec['stop_splunk'],
+      require     => [Exec['stop_splunk'], Exec['disable_splunk_boot_start']],
+      subscribe   => Package[$splunk::enterprise::package_name],
```
bastelfreak, line 38: I think the `subscribe` isn't required because the new exec already subscribes to the same package resource?

manifests/enterprise/service/nix.pp
siegy22: Sure! I just used the same the string for local consistency.

manifests/enterprise/service/nix.pp
siegy22: Actually it is needed, I just tested this using this snippet:

```puppet
file { '/tmp/foo':
  content => "foo",
}

exec { 'bar':
  command => "/usr/bin/echo bar",
  refreshonly => true,
  subscribe => File['/tmp/foo'],
}

exec { 'baz':
  command => "/usr/bin/echo baz",
  refreshonly => true,
  require => Exec['bar']
}
```

In this scenario, `baz` was never executed.

bastelfreak: ah right. Because `require` only enforces the ordering, but it doesn't do any refresh. 

```puppet
exec { 'enable_splunk':
    command     => "${splunk::enterprise::enterprise_homedir}/bin/splunk enable boot-start ${user_args} ${splunk::params::boot_start_args} --accept-license --answer-yes --no-prompt",
    refreshonly => true,
    before      => Service[$splunk::enterprise::service_name],
    subscribe   => Exec['disable_splunk_boot_start']
}
```

This is probably enough?

* `Exec['stop_splunk']` notifies  `Exec['enable_splunk']`, so `Exec['enable_splunk']` doesn't need to require `Exec['stop_splunk']`
* `Exec['disable_splunk_boot_start']` is already subscribed to `Package[$splunk::enterprise::package_name]`, so `Exec['enable_splunk']`  doesn't need to subscribe to `Package[$splunk::enterprise::package_name]`
* `Exec['enable_splunk']` needs to subscribe to `Exec['disable_splunk_boot_start']` for proper ordering

siegy22: I guess it kind of does the same, before it was done by subscribing to the Package, and then setting the order by adding it to the `require` parameter. 
But I changed it to the way you proposed it 😄

siegy22:
@bastelfreak Ready to merge? 🎉

PR #371, opened 2024-04-01
Treat the forwarder user as a separate parameter

### LINKED ISSUE #369: Wrong Splunk Forwarder user on Windows

- Puppet: Any
- Ruby: Any
- Distribution: Windows (any modern version)
- Module version: 10.0.0

include puppet::forwarder

## What are you seeing
The Linux package uses `splunkfwd` as the default user and group names. On Windows, `SplunkForwarder` is used for the user and group names. This causes the module to fail setting file permissions.

## Output log
```
change from 'NT AUTHORITY\SYSTEM' to 'splunkfwd' failed: Could not find user splunkfwd

    Source: /Stage[main]/Splunk::Forwarder::Config/File[C:\Program Files\SplunkUniversalForwarder/etc/system/local/server.conf]/owner 
```

Tamerz:
Now the forwarder uses a completely different user name. I thought using a new `splunk::params::splunk_forwarder_user` would make things cleaner.

It also sets the forwarder user to the correct default user for Windows.

Fixes #369

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -994,13 +994,7 @@ Data type: `String[1]`
 
 The user to run Splunk as.
 
-Default value:
-
-```puppet
-versioncmp($version, '9.1.0') ? {
-    -1 => $splunk::params::splunk_user,
-    default => 'splunkfwd'
-```
+Default value: `$splunk::params::splunk_forwarder_user`
 
 ##### <a name="-splunk--forwarder--forwarder_homedir"></a>`forwarder_homedir`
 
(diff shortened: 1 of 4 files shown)


kenyon (commented):
Converting `params.pp` to hiera data would simplify this, but that's a separate task. LGTM once `REFERENCE.md` is updated so that the tests can run. `bundle exec rake strings:generate:reference`

Also, please rebase on `master` to remove the merge commit. Thanks.

PR #378, opened 2024-08-20
Correct `splunk.secret` path

### LINKED ISSUE #324: default location of splunk.secret is wrong?
- Puppet: 6.24
- Ruby:
- Distribution: Debian 10
- Module version: 9.0.0
- Splunk version: 8.1.0

## What are you seeing
I noticed that in params.pp the secret file location is set to '${homedir}/etc/splunk.secret'.

In our instances the splunk.secret file is in ${homedir}/etc/auth/. Also in splunk::enterprise::service::nix the /etc/auth/splunk.secret is used as hard-coded location for the exec creates.

I was wondering if this default is correct, and, if yes, which magic I'm missing here. If the default location is indeed incorrect I can supply a patch for this.

luedderd:

The `splunk.secret` file on Linux is located at `$SPLUNK_HOME/etc/auth` instead of `$SPLUNK_HOME/etc/` as per [Splunk documentation](https://docs.splunk.com/Documentation/Splunk/9.3.0/Security/Deploysecurepasswordsacrossmultipleservers) since at least v7.0.0 through v9.3.0

Fixes #324

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -159,8 +159,8 @@ class splunk::params (
       $enterprise_seed_config_file     = "${enterprise_homedir}/etc/system/local/user-seed.conf"
       $forwarder_password_config_file  = "${forwarder_homedir}/etc/passwd"
       $enterprise_password_config_file = "${enterprise_homedir}/etc/passwd"
-      $forwarder_secret_file           = "${forwarder_homedir}/etc/splunk.secret"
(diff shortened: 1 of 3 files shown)


kenyon (changes requested):
Can you also make this change for Windows? I can confirm it's under `etc/auth` on my Windows system. It's likely the same for all of the OSes.

kenyon (approved):

PR #389, opened 2024-11-11
Adding Edge Processor support

### LINKED ISSUE #388: Edge Processor Support
Seems reasonable that we'd want to add support for installing/configuring the Edge Processor via this module. 

Biggest difference off the bat is that the Edge Processor is installed simply by unpacking a tar.gz file, not installing it via a package manager. Please give me some feedback on how I've decided to handle this, as there's probably a better way to do this. 

I'm modelling this on the existing structure of the Forwarder classes, but I'm not an expert Puppet developer by any means so I welcome any and all constructive feedback on how to improve this addition. 

I've started work at https://github.com/luedderd/puppet-splunk and I'll open a PR once I'm done testing on my side, but I wanted to open this to get some eyes on it. Thanks!

luedderd:
Adding support for installing/configuring/managing the Edge Processor from Splunk.

Fixes #388

### DIFF
--- a/lib/puppet/type/splunk_config.rb
+++ b/lib/puppet/type/splunk_config.rb
@@ -22,6 +22,12 @@ Puppet::Type.newtype(:splunk_config) do
 
   newparam(:server_confdir) do
   end
+
+  newparam(:edgeproc_installdir) do
+  end
+
+  newparam(:edgeproc_confdir) do
+  end
   # rubocop:enable Lint/EmptyBlock
 
   ## Generate purge parameters for the splunk_config type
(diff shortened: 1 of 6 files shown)


luedderd:
Ok another annoying note is that the authentication token used here to connect the Edge Processor back to SCS expires after 5 minutes, so after initial deployment, you'll have to update the token again to deploy to additional servers or if they go offline for long enough. This feels like a design issue on Splunk's side, and I'm not sure if/how we want to handle that, or that kind of inflexibility invalidates the worth of using Puppet to manage/deploy Edge Processors. Thoughts?

siegy22:
Can you update the REFERENCE.md file by running `bundle exec rake strings:generate:reference`? 
Then the actual test suite should run with all the version and OS matrix builds.

PR #424, opened 2026-04-17
Fix ownership and upgrade handling for Splunk Forwarder

deepak7340:
- Add upgrade logic in install.pp to detect version changes and handle stop/disablebootstart/install/enablebootstart/fixownership
- Add -group to boot-start in service/nix.pp to set correct group
- Fix ownership of TA app files after archive extraction in addon.pp
- Add systemd symlink fix after boot-start enable in service/nix.pp

---
*PR sponsored by [Obmondo.com](https://obmondo.com)*

---

**Note:** Since this is a company-paid contribution, we fork to the company
account repo on GitHub. GitHub does not support "allow maintainer to edit
in repo" for organization accounts.

- Maintainers can either provide feedback for required changes before merging, OR
- Pull our branch, make changes in their repo, and create a new PR including our commits.

### DIFF
--- a/manifests/forwarder/service/nix.pp
+++ b/manifests/forwarder/service/nix.pp
@@ -19,7 +19,12 @@ class splunk::forwarder::service::nix inherits splunk::forwarder::service {
       notify  => Exec['enable_splunkforwarder'],
     }
 
-    $user_args = "-user ${splunk::forwarder::splunk_user}"
+    $user_args = "-user ${splunk::forwarder::splunk_user} -group ${splunk::forwarder::splunk_user}"
+
+    $user_args_full = $splunk::params::boot_start_args ? {
+      ''      => $user_args,
+      default => "${user_args} ${splunk::params::boot_start_args}",
+    }
 
     if $facts['kernel'] == 'SunOS' {
       Service[$splunk::forwarder::service_name] {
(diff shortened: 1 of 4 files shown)


manifests/forwarder/service/nix.pp
```diff
@@ -37,6 +37,13 @@
       before      => Service[$splunk::forwarder::service_name],
       require     => Exec['stop_splunkforwarder'],
     }
+
+    exec { 'fix-splunkforwarder-systemd-symlink':
+      command     => '/bin/cp /usr/lib/systemd/system/SplunkForwarder.service /etc/systemd/system/SplunkForwarder.service',
+      onlyif      => '/usr/bin/test -L /etc/systemd/system/SplunkForwarder.service',
+      refreshonly => true,
+      subscribe   => Exec['enable_splunkforwarder'],
+    }
```
kenyon: Why copy the service unit file from `/usr` to `/etc`?

deepak7340: While testing the code on the SUSE server, we encountered a symlink issue:

`/etc/systemd/system/SplunkForwarder.service -> /usr/lib/systemd/system/SplunkForwarder.service`

This setup leads to errors when systemd attempts to verify the service state:

`Failed to get unit file state for SplunkForwarder.service: Too many levels of symbolic links`

As a result, Puppet’s systemctl enable check fails. To resolve this, the actual service file should be copied into /etc/systemd/system/ instead of relying on the symlink. This allows systemd to correctly read and manage the service.

kenyon: What creates that symlink?

I think the right thing to do is remove `/etc/systemd/system/SplunkForwarder.service` and only use the file from `/usr`. But I'm not sure that should be done in this module, since users might have placed a customized file into `/etc` for some reason.

deepak7340: You’re right—we shouldn’t modify files under /etc since they may be user-customized. I’ve removed the symlink-copying fix from the PR. The final version no longer includes any copy or removal logic for the systemd service file; instead, it relies on Splunk’s boot-start command to create it correctly using the -group parameter.

deepak7340: @kenyon  can you merge this

